PR #11, opened 2017-03-06
core: cleanups, in preparation for Windows port

mslusarz:
This change is [Reviewable](https://reviewable.io/reviews/marcinslusarz/pmemfile/11)


src/libpmemfile-core/dir.c
```diff
@@ -34,20 +34,20 @@
...
  */
 
-#include <dirent.h>
 #include <errno.h>
 #include <limits.h>
 #include <stdio.h>
 
 #include "callbacks.h"
 #include "dir.h"
+
```
GBuella: additional empty line needed?

src/libpmemfile-core/inode.c
```diff
@@ -34,24 +34,19 @@
...
-#include <stdlib.h>
 #include <unistd.h>
-#include <sys/stat.h>
-#include <sys/types.h>
 
 #include "callbacks.h"
 #include "data.h"
 #include "dir.h"
 #include "inode.h"
+
```
GBuella: Is this empty line needed?

src/libpmemfile-core/os_thread.h
```diff
@@ -0,0 +1,124 @@
...
+#define PMEMFILE_OS_LOCKS_H 1
+
+typedef union {
+	long long align;
+	char padding[64];
+} os_mutex_t;
+
+typedef union {
+	long long align;
+	char padding[64];
```
GBuella: Are these used internally only?
And in volatile memory?
I suspect yes.
Then please, don't create such a placeholder type, no need for specifying 64 byte size.
Please just typedef the mutex available on the platform, as in:
https://github.com/tinycthread/tinycthread/blob/master/source/tinycthread.h#L194

The 64 byte assumption is just plain wrong, there is no standard anywhere, that says this thing should fit in 64 bytes. There is no such magic number.

E.g. it might as well be 200 bytes:
https://github.com/pmem/issues/issues/427

marcinslusarz: These are used only internally and only for volatile memory. I explicitly didn't want to typedef to pthread_mutex_t, because it requires inclusion of pthread.h, which means anything imported by pthread.h will leak to the code. 
I was perfectly aware of pmem/issues#427 - when porting to such platform we'll just extend the size of this union. It's not part of API or ABI, so we are free to do this at any time.

In os_locks_pthread.c there is this code:
COMPILE_ERROR_ON(sizeof(os_mutex_t) < sizeof(pthread_mutex_t));
which will catch platforms where 64 byte assumption is wrong.

src/libpmemfile-core/inode.c
marcinslusarz: nope, will remove

src/libpmemfile-core/dir.c
marcinslusarz: nope, will remove

src/libpmemfile-core/os_thread.h
GBuella: Well, at least it could be simpler:
typedef struct {
   long long data[8];
} os_mutext_t;

marcinslusarz: Sure. I'll do that.

PR #15, opened 2017-03-06
tests: fix build with non-default compiler

marcinslusarz:
Without this change, when user sets CC/CXX vars before cmake, they do
not propagate to gtest, which means it's built with default compiler.
This can lead to linking error when the difference between compilers
is big enough.

Ref: marcinslusarz/pmemfile#13

---
This change is [Reviewable](https://reviewable.io/reviews/marcinslusarz/pmemfile/15)


GBuella:
:lgtm:

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/marcinslusarz/pmemfile/15#-:-KeYwG9XrzQzvPKTxzGa:bnfp4nl)*

PR #20, opened 2017-03-07
core: remove or mark unused arguments

GBuella:
Unused arguments removed in the case of static
functions (unless they are used as callback functions),
marked as known unused arguments otherwise.

---
This change is [Reviewable](https://reviewable.io/reviews/marcinslusarz/pmemfile/20)

### DIFF
--- a/src/libpmemfile-core/inode.c
+++ b/src/libpmemfile-core/inode.c
@@ -80,6 +80,8 @@ pmfi_path(struct pmemfile_vinode *vinode)
 struct pmemfile_vinode *
 vinode_ref(PMEMfilepool *pfp, struct pmemfile_vinode *vinode)
 {
+	(void) pfp;
+
 	__sync_fetch_and_add(&vinode->ref, 1);
 	return vinode;
 }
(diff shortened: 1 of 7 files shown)


src/libpmemfile-core/inode.c
```diff
@@ -80,6 +80,8 @@ pmfi_path(struct pmemfile_vinode *vinode)
 struct pmemfile_vinode *
 vinode_ref(PMEMfilepool *pfp, struct pmemfile_vinode *vinode)
 {
+	(void) pfp; /* XXX: could this ever be needed? */
```
marcinslusarz: Maybe for logging / debugging? I don't know. I left these pfp's everywhere just in case they will be needed...

GBuella: OK, so do you suggest I remove this comment?
Or LGTY?

marcinslusarz: Yes, please remove this comment (and the one in inode_free). Other than that it LGTM.

GBuella: Done.

PR #22, opened 2017-03-08
core: refactor block_array handling

GBuella:
Introduce routines allocate_new_block_metadata and
unlink_and_release_block_metadata.

---
This change is [Reviewable](https://reviewable.io/reviews/marcinslusarz/pmemfile/22)

### DIFF
--- a/src/libpmemfile-core/block_array.c
+++ b/src/libpmemfile-core/block_array.c
@@ -1,0 +1,230 @@
+/*
+ * Copyright 2016-2017, Intel Corporation
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ *
+ *     * Redistributions of source code must retain the above copyright
+ *       notice, this list of conditions and the following disclaimer.
+ *
+ *     * Redistributions in binary form must reproduce the above copyright
+ *       notice, this list of conditions and the following disclaimer in
+ *       the documentation and/or other materials provided with the
+ *       distribution.
+ *
+ *     * Neither the name of the copyright holder nor the names of its
+ *       contributors may be used to endorse or promote products derived
+ *       from this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+ * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include "layout.h"
+#include "inode.h"
+#include "internal.h"
+#include "out.h"
+
+static void
+update_binfo(struct pmemfile_vinode *vinode)
+{
+	struct pmemfile_inode *inode = D_RW(vinode->inode);
+	struct block_info *binfo = &vinode->first_free_block;
+
+	if (binfo->arr != NULL) {
+		/*
+		 * If the block_info is not null, it means it was kept
+		 * up-to-date by the allocating routines.
+		 */
+		return;
+	}
+
+	/*
+	 * If binfo was not used before, it must be initialized.
+	 */
+
+	/*
+	 * Find the block_array containing the next free block metadata
+	 * slot. This is either the block_array stored right in the
+	 * inode, ...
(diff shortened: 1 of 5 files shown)


src/libpmemfile-core/block_array.c
```diff
@@ -0,0 +1,230 @@
...
+		binfo->idx++;
+}
+
+static bool
+has_free_block_entry(struct pmemfile_vinode *vinode)
+{
+	struct block_info *binfo = &vinode->first_free_block;
+
+	return binfo->idx < binfo->arr->length;
+}
```
marcinslusarz: After previous acquire_new_entry binfo->idx may point to used entry. Previous code handled it correctly.

src/libpmemfile-core/layout.h
```diff
@@ -191,4 +191,10 @@ struct pmemfile_super {
 			- 16 /* toid */];
 };
 
+static inline TOID(struct pmemfile_block)
+blockp_as_oid(struct pmemfile_block *block)
+{
+	return (TOID(struct pmemfile_block))pmemobj_oid(block);
+}
+
```
marcinslusarz: Please don't add any code to layout.h.

src/libpmemfile-core/block_array.c
```diff
@@ -0,0 +1,230 @@
...
+
+	return binfo->arr->blocks + (binfo->idx - 1);
+}
+
+static void
+unlink_block(struct pmemfile_block *block)
+{
+	if (!TOID_IS_NULL(block->prev)) {
+		TX_ADD_FIELD_DIRECT(D_RW(block->prev), next);
+		D_RW(block->prev)->next = block->next;
```
marcinslusarz: Please cache the result of D_RW

src/libpmemfile-core/block_array.c
```diff
@@ -0,0 +1,230 @@
...
+	}
+
+	if (!TOID_IS_NULL(block->next)) {
+		TX_ADD_FIELD_DIRECT(D_RW(block->next), prev);
+		D_RW(block->next)->prev = block->prev;
```
marcinslusarz: And here.

src/libpmemfile-core/block_array.c
```diff
@@ -0,0 +1,230 @@
...
+static void
+relocate_block(struct pmemfile_block *dst, struct pmemfile_block *src)
+{
+	ASSERT(dst != src);
+
+	TX_ADD_DIRECT(dst);
+
+	if (!TOID_IS_NULL(src->prev)) {
+		TX_ADD_FIELD_DIRECT(D_RW(src->prev), next);
+		D_RW(src->prev)->next = blockp_as_oid(dst);
```
marcinslusarz: .

src/libpmemfile-core/block_array.c
```diff
@@ -0,0 +1,230 @@
...
+	}
+
+	if (!TOID_IS_NULL(src->next)) {
+		TX_ADD_FIELD_DIRECT(D_RW(src->next), prev);
+		D_RW(src->next)->prev = blockp_as_oid(dst);
```
marcinslusarz: .

src/libpmemfile-core/block_array.c
```diff
@@ -0,0 +1,230 @@
...
+	TOID(struct pmemfile_block_array) new_next;
+	struct block_info *binfo;
+
+	inode = D_RW(vinode->inode);
+	binfo = &vinode->first_free_block;
+
+	to_remove = inode->file_data.blocks.next;
+
+	new_next = D_RW(to_remove)->next;
+	inode->file_data.blocks.next = new_next;
```
marcinslusarz: snapshotting is missing

src/libpmemfile-core/block_array.c
```diff
@@ -0,0 +1,230 @@
...
+		binfo->arr = &inode->file_data.blocks;
+	else
+		binfo->arr = D_RW(new_next);
+
+	TX_FREE(to_remove);
+	binfo->idx = binfo->arr->length;
+}
+
+void
+unlink_and_release_block_metadata(struct pmemfile_vinode *vinode,
```
marcinslusarz: Maybe just "free_block"?

src/libpmemfile-core/block_array.c
```diff
@@ -0,0 +1,230 @@
...
+		struct pmemfile_block *block,
+		struct pmemfile_block **prev)
```
marcinslusarz: I think returning this would look better.

src/libpmemfile-core/block_array.c
GBuella: why?

src/libpmemfile-core/block_array.c
GBuella: well, the allocate counter part doesn't link the prev and next fields, but this one does take care of those fields.
So, maybe unlink_and_free_block?

src/libpmemfile-core/block_array.c
marcinslusarz: Because it's useless work.

In this case I think you could replace these 2 lines with other 2 lines:
struct pmemfile_block *prev = D_RW(block->prev);
TX_SET_DIRECT(prev, next, block->next);

or even just one line:
TX_SET(block->prev, next, block->next);

src/libpmemfile-core/block_array.c
marcinslusarz: A function with "and" in the name is a red flag to me. It either does two things and should be split or does one thing and its purpose is not well understood ;).

GBuella: Well, it does two things in a sense, but one thing in another sense.
I think I'll pick the second one.
Going to refactor it a little more:
There is a doubly linked list of blocks, and we can implement an 'insert' and a 'remove' operation for such a list.
The unlink_and_release_block_metadata corresponds to 'remove'-ing from the list.
By the same convention, the 'insert' operation should both allocate, and modify the appropriate prev/next pointers - thus link it into the linked list.
Currently this linking happens in data.c in allocate_range. So allocate_range should just call something like 'insert_after', and never modify the prev/next fields itself.

That should make things clear.

src/libpmemfile-core/block_array.c
GBuella: Yes, that is one way to do it.
I was worried, that a return value has no name attached to it.
So, what does a 'remove' operation return? Would you expect it to return the address of the previous node in the list?

marcinslusarz: Just make a comment describing what this function returns?

GBuella:
Going to rework this, and make another PR

PR #32, opened 2017-03-13
Refactor block array usage

GBuella:
This change is [Reviewable](https://reviewable.io/reviews/marcinslusarz/pmemfile/32)


GBuella:
Going to make a bit smaller PR instead of this

PR #38, opened 2017-03-14
posix: port to Windows, part 3

mslusarz:
This change is [Reviewable](https://reviewable.io/reviews/marcinslusarz/pmemfile/38)


src/util.h
```diff
@@ -67,6 +67,14 @@ extern "C" {
 
 #define COMPILE_ERROR_ON(cond) ((void)sizeof(char[(cond) ? -1 : 1]))
 
+#define pf_always_inline __attribute__((always_inline)) inline
```
GBuella: Are these really needed?
I would call it useless obfuscation.

tests/core/permissions/permissions.cpp
```diff
@@ -100,7 +100,9 @@ TEST_F(permissions, chmod)
 		SCOPED_TRACE(m);
 
 		/* chmod u+r */
-		ASSERT_EQ(pmemfile_chmod(pfp, "/aaa", S_IRUSR | (mode_t)m), 0)
+		ASSERT_EQ(pmemfile_chmod(pfp, "/aaa",
+					 PMEMFILE_S_IRUSR | (mode_t)m),
+			  0)
```
GBuella: Indentation related typo.

src/libpmemfile-core/dir.c
```diff
@@ -102,6 +102,17 @@ component_length(const char *path)
 	return (uintptr_t)slash - (uintptr_t)path;
 }
 
+#ifdef DEBUG
+static inline char *
+util_strndup(const char *c, size_t len)
+{
+	char *cp = malloc(len);
+	memcpy(cp, c, len - 1);
```
GBuella: strncpy might fit better here.

src/libpmemfile-core/out.c
```diff
@@ -39,11 +39,20 @@
...
 #include <limits.h>
 #include <string.h>
 #include <errno.h>
 
+#ifdef DEBUG
+#ifndef _WIN32
+#include <unistd.h>
+#define os_getpid getpid
+#else
+#include <process.h>
```
GBuella: Well, if have the os_thread things in a separate TU, I would prefer to do the same with this as well. Who knows what macros are defined in process.h
Perhaps we could add a os_util.h, os_util.c ?

src/util.h
marcinslusarz: They are not needed now. If we'll ever port pmemfile to MSVC these will be needed. And I think having wrappers around compiler-specific code like attributes is a "good thing" (tm).

tests/core/permissions/permissions.cpp
marcinslusarz: What do you mean? It's how clang-format formats this code.

src/libpmemfile-core/dir.c
mslusarz: Well, in the only place it's needed, the string we want to copy is exactly len bytes long, so using strncpy will do useless NUL-termination checks for each byte.

FTR, this implementation has a off-by-one bug. I'll push a new version in few minutes.

src/libpmemfile-core/out.c
mslusarz: Done

src/util.h
GBuella: One other way I use sometimes, is:

#ifdef HAS_GNU_ATTR
#define attribute __attribute__
#define attribute(...) __attribute__((__VA_ARGS__))
#else
#define attribute(...)
#endif

Or just define this macro in CFLAGS

tests/core/permissions/permissions.cpp
GBuella: OK, I see now.

src/libpmemfile-core/dir.c
GBuella: OK

src/util.h
marcinslusarz: The point of these macros is to define them in a way that is as far from compiler as possible, because on different compiler the name of an attribute can be different or the syntax will look differently.

And even in this case it can't work - there's no way to include "inline" in your definition of always_inline attribute.

GBuella: Thats on of my problems with it, from the other macros, one would expect pf_always_inline defines \_\_attribute\_\_((always_inline))

But it is an exception, it also the the inline keyword.
It is obfuscation.

PR #40, opened 2017-03-15
Integrate AdressSanitizer/UndefinedBehaviorSantitizer and fix some issues

marcinslusarz:
This change is [Reviewable](https://reviewable.io/reviews/marcinslusarz/pmemfile/40)


marcinslusarz:
Some posix tests still fail with ASAN, but they seem to be false positives.
Preload tests don't work with ASAN at all and I don't know how to fix it yet.

GBuella:
For now, we might need to make wish to allow this only when BUILD_LIBPMEMFILE is set to OFF .
Otherwise, issue a meaningful error message.

marcinslusarz:
Done.

PR #44, opened 2017-03-17
[RFC] posix: port to Windows, part 4

mslusarz:
On Windows there are problems with types.

Some typedefs don't exist (like ssize_t) and some are wrong (like off_t being always 32-bit and off64_t not existing). The only portable way seems to be to define our own typedefs, use them everywhere and on Linux validate that they have the same size as their native counterparts.

Another problem is with "struct stat". It's defined on Windows, but it has less fields (no st_blksize and st_blocks) and fields that exist have shorter types (eg st_size is 32-bit). There's "struct \_stat64" with some of the fields bigger (st_size is 64-bit), but still some of them are too small (st_uid, st_gid, st_*time, nlink_t, st_ino). The solution presented in this PR is to define "struct pmemfile_stat" exactly like on Linux, use it and validate its size and layout matches native struct stat.

---
This change is [Reviewable](https://reviewable.io/reviews/marcinslusarz/pmemfile/44)

### DIFF
--- a/include/libpmemfile-posix.h
+++ b/include/libpmemfile-posix.h
@@ -207,11 +207,6 @@ ssize_t pmemfile_read(PMEMfilepool *pfp, PMEMfile *file, void *buf,
 off_t pmemfile_lseek(PMEMfilepool *pfp, PMEMfile *file, off_t offset,
 		int whence);
 
-#ifdef __off64_t_defined
-off64_t pmemfile_lseek64(PMEMfilepool *pfp, PMEMfile *file, off64_t offset,
-		int whence);
-#endif
-
 ssize_t pmemfile_pwrite(PMEMfilepool *pfp, PMEMfile *file, const void *buf,
 		size_t count, off_t offset);
 
--- a/include/libpmemfile-posix.h
+++ b/include/libpmemfile-posix.h
@@ -170,9 +170,24 @@ extern "C" {
 typedef struct pmemfilepool PMEMfilepool;
 typedef struct pmemfile_file PMEMfile;
 
+typedef mode_t pmemfile_mode_t;
+typedef uid_t pmemfile_uid_t;
+typedef gid_t pmemfile_gid_t;
+typedef ssize_t pmemfile_ssize_t;
+#ifndef __off64_t_defined
+#error off64_t is not available, on glibc enable _LARGEFILE64_SOURCE, _XOPEN_SOURCE>=500 or _GNU_SOURCE to define it
+#endif
+typedef off64_t pmemfile_off_t;
+typedef nlink_t pmemfile_nlink_t;
+typedef blksize_t pmemfile_blksize_t;
+typedef blkcnt64_t pmemfile_blkcnt_t;
+typedef dev_t pmemfile_dev_t;
+typedef ino64_t pmemfile_ino_t;
+
 #define PMEMFILE_AT_CWD ((PMEMfile *)(((unsigned char *)0) - 1))
 
-PMEMfilepool *pmemfile_mkfs(const char *pathname, size_t poolsize, mode_t mode);
+PMEMfilepool *pmemfile_mkfs(const char *pathname, size_t poolsize,
+		pmemfile_mode_t mode);
 
 PMEMfilepool *pmemfile_pool_open(const char *pathname);
 void pmemfile_pool_close(PMEMfilepool *pfp);
@@ -181,7 +196,8 @@ PMEMfile *pmemfile_open(PMEMfilepool *pfp, const char *pathname, int flags,
 		...);
 PMEMfile *pmemfile_openat(PMEMfilepool *pfp, PMEMfile *dir,
 		const char *pathname, int flags, ...);
-PMEMfile *pmemfile_create(PMEMfilepool *pfp, const char *pathname, mode_t mode);
+PMEMfile *pmemfile_create(PMEMfilepool *pfp, const char *pathname,
+		pmemfile_mode_t mode);
 /* XXX Should we get rid of PMEMfilepool pointer? */
 void pmemfile_close(PMEMfilepool *pfp, PMEMfile *file);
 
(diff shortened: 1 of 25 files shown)


include/libpmemfile-posix.h
```diff
@@ -158,9 +157,47 @@ extern "C" {
...
+typedef unsigned pmemfile_gid_t;
+typedef long long pmemfile_ssize_t;
+typedef long long pmemfile_off_t;
+typedef unsigned long pmemfile_nlink_t;
+typedef long pmemfile_blksize_t;
+typedef long pmemfile_blkcnt_t;
+typedef unsigned long pmemfile_dev_t;
+typedef unsigned long pmemfile_ino_t;
+
+struct pmemfile_timespec
```
GBuella: The struct timespec is in ISO C.
Naturally can't expect POSIX on Windows, but I would prefer to keep the types described in ISO.
Does it really seem necessary?

GBuella: Oops, I realized it is only in C11, but not in C89 or C99.

src/libpmemfile-core/libpmemfile-posix.map
```diff
@@ -68,7 +68,6 @@
 		pmemfile_link;
 		pmemfile_linkat;
 		pmemfile_lseek;
-		pmemfile_lseek64;
```
GBuella: Reminder:
If merged, don't forget to remove this from the man page as well.

src/libpmemfile/path_resolve.c
```diff
@@ -69,7 +69,7 @@ get_stat(struct resolved_path *result, struct stat *buf)
 		int r = pmemfile_fstatat(result->at.pmem_fda.pool->pool,
 			result->at.pmem_fda.file,
 			result->path,
-			buf,
+			(struct pmemfile_stat *)buf,
```
GBuella: Or just change the get_stat function's signature to accept pmemfile_stat.
The syscall_no_intercept function doesn't care about types anyways.

include/libpmemfile-posix.h
```diff
@@ -158,9 +157,47 @@ extern "C" {
...
+{
+	long tv_sec;
+	long tv_nsec;
+};
+
+struct pmemfile_stat
+{
+	pmemfile_dev_t st_dev;
+	pmemfile_ino_t st_ino;
+	pmemfile_nlink_t st_nlink;
```
GBuella: Is this typedef justified now?
Can't we just say:
'unsigned st_nlink' ?
Since these all are just integers, the compiler doesn't really catch errors where an argument of wrong type is passed to a function. Theoretically, 'unsigned' only guarantees an at least 16 bits wide integer, and what if a file has 70000 links...
Anyways, perhaps we don't have to prepare for any exotic scenario that POSIX prepares for.

include/libpmemfile-posix.h
```diff
@@ -158,9 +157,47 @@ extern "C" {
...
+	pmemfile_mode_t st_mode;
+	pmemfile_uid_t st_uid;
+	pmemfile_gid_t st_gid;
+	int __pad0;
+	pmemfile_dev_t st_rdev;
+	pmemfile_off_t st_size;
+	pmemfile_blksize_t st_blksize;
+	pmemfile_blkcnt_t st_blocks;
```
GBuella: Similarly, I don't think a typedef is needed for the 'count of something'.
I think 'long st_blocks' pretty much expresses, that it is a non-negative integer.
I don't much like the other typedefs either.

GBuella:
A few comments on the idea of "exactly like on Linux"

Thinking about it, one better approach might be something like:
```
#ifdef ON_POSIX_LIKE_PLATFORM

#include <sys/stat.h>

typedef mode_t pmemfile_mode_t;
typedef stat pmemfile_stat;
// etc..

#else

struct pmemfile_stat {
};

#endif
```

Or, rather just create two versions of the public headers, instead of using preprocessor conditions.

But only, if we really expect users to do such things as:
```
mode_t a;
pmemfile_mode_t b;


a = b; // no warning triggered, even if compiler would warn about 'long' != 'long long' if both are 64 bits wide

b = a; // ditto
```

I'm not sure such total compatibility is needed, after all, people will have to rewrite their code anyways to use the pmemfile-posix API. Especially if used intertwined with the good-old POSIX API -- lot of code needs to be developed for that.

But if possible, the solution I would prefer the most, is something like:
```
#ifdef ON_POSIX_LIKE_PLATFORM

#include <sys/stat.h>

typedef mode_t pmemfile_mode_t;
typedef stat pmemfile_stat;
// etc..

#else

struct pmemfile_stat {
  unsigned st_mode;
  unsigned st_blocks;
  // etc...
};

#endif
```

With this approach, we wouldn't need to use those COMPILE_ERROR_ON batches.
I mean, look at it:
1) Manually define type A, expecting it to be equivalent to type B
2) Verifying that type A is equivalent to type B

Doesn't this look much more sane:
1) Defining type A to be an alias for type B
?

And if ever, in the unforeseeable future this is ported to other POSIX-like platforms, we won't have to manually keep track of these.

If anything, the compile time assertions should check these types against what is supported in the binary layout of pmemfile, e.g.:
If the number of links supported is minimum 0 and maximum 10000, then a compile time assertion should check that pmemfile.st_nlink can store such numbers.

sarahjelinek:
What did they do for the windows port of the NVML libraries?

marcinslusarz:
NVML's API is almost posix-independent. The only thing that was missing from API standpoint was mode_t (used in _create functions) and struct iovec (used only in pmemlog API). What happened is they included pmemcompat.h under ifdef _WIN32 in all exported headers and defined mode_t and struct iovec in pmemcompat.h.
This is wrong, as those types can be defined by other such "compat" headers.

src/libpmemfile-core/libpmemfile-posix.map
mslusarz: Fixed. For now only from markdown.

src/libpmemfile/path_resolve.c
mslusarz: Then get_stat user will have to use pmemfile_stat, which means more code will have to use it and I think it doesn't make much sense.

include/libpmemfile-posix.h
mslusarz: If we change the type of this field to unsigned, then pmemfile_stat size and layout won't match native Linux struct stat (and pmemfile_inode already has nlink of type uint64_t, so...).
Assuming you meant unsigned long long (which I just realized pmemfile_nlink_t should be instead of current "unsigned long"), I think we should have pmemfile versions of all types if we want applications using pmemfile-posix to be portable.

include/libpmemfile-posix.h
mslusarz: The thing is blkcnt_t and blksize_t are signed. We are just providing portable types.

mslusarz:
Gabor, I tried your proposed approach (see last commit), but I don't like it. The code that wants to print the variables which are defined as long on Linux and long long on Windows in portable way needs to cast them to long long. Also there's a problem with off64_t... :/

mslusarz:
Ok, as agreed I defined pmemfile_ typedefs using native types. For now I removed all Windows-specific code - I'll propose it as a separate PR once I figure out how to add it in the cleanest possible way.

Unfortunately with this approach it's not possible to discover uses of mode_t (instead of pmemfile_mode_t) without building it on Windows. Oh well...

PR #50, opened 2017-03-20
posix: add pmemfile_setcap, pmemfile_clrcap and use it in chmod

marcinslusarz:
This change is [Reviewable](https://reviewable.io/reviews/marcinslusarz/pmemfile/50)

PR #56, opened 2017-03-21
posix: implement chown, fchown, lchown

marcinslusarz:
Built on top of PR #50

---
This change is [Reviewable](https://reviewable.io/reviews/marcinslusarz/pmemfile/56)

### DIFF
--- a/src/libpmemfile-core/file.c
+++ b/src/libpmemfile-core/file.c
@@ -1674,3 +1674,171 @@ end:
 
 	return 0;
 }
+
+static int
+vinode_chown(PMEMfilepool *pfp, struct pmemfile_cred *cred,
+		struct pmemfile_vinode *vinode, uid_t owner, gid_t group)
+{
+	struct pmemfile_inode *inode = vinode->inode;
+	int error = 0;
+
+	if (owner == (uid_t)-1 && group == (gid_t)-1)
+		return 0;
+
+	os_rwlock_wrlock(&vinode->rwlock);
+
+	if (!(cred->caps & (1 << PMEMFILE_CAP_CHOWN))) {
+		if (inode->uid != cred->fsuid) {
+			error = EPERM;
+			goto end;
+		}
+
+		if (owner != (uid_t)-1 && owner != inode->uid) {
+			error = EPERM;
+			goto end;
+		}
+
+		if (group != (gid_t)-1 && group != inode->gid) {
+			if (group != cred->fsgid && !gid_in_list(cred, group)) {
+				error = EPERM;
+				goto end;
+			}
+		}
+	}
+
+	TX_BEGIN_CB(pfp->pop, cb_queue, pfp) {
+		COMPILE_ERROR_ON(offsetof(struct pmemfile_inode, gid) !=
+				offsetof(struct pmemfile_inode, uid) +
+				sizeof(inode->uid));
+
+		pmemobj_tx_add_range_direct(&inode->uid,
+				sizeof(inode->uid) + sizeof(inode->gid));
+
+		if (owner != (uid_t)-1)
+			inode->uid = owner;
+		if (group != (gid_t)-1)
+			inode->gid = group;
+	} TX_ONABORT {
+		error = errno;
+	} TX_END
+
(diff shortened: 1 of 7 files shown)


src/libpmemfile-core/file.c
```diff
@@ -1674,3 +1674,158 @@ pmemfile_truncate(PMEMfilepool *pfp, const char *path, off_t length)
...
+					!gid_in_list(cred, group))
+				pmemfile_tx_abort(EPERM);
+		}
+
+		COMPILE_ERROR_ON(offsetof(struct pmemfile_inode, gid) !=
+				offsetof(struct pmemfile_inode, uid) +
+				sizeof(inode->uid));
+
+		pmemobj_tx_add_range_direct(&inode->uid,
+				sizeof(inode->uid) + sizeof(inode->gid));
```
GBuella: I don't think we need to care a lot about chown being 'fast', but whatever...

src/libpmemfile-core/file.c
```diff
@@ -1674,3 +1674,158 @@ pmemfile_truncate(PMEMfilepool *pfp, const char *path, off_t length)
...
+vinode_chown(PMEMfilepool *pfp, struct pmemfile_cred *cred,
+		struct pmemfile_vinode *vinode, uid_t owner, gid_t group)
+{
+	struct pmemfile_inode *inode = vinode->inode;
+	int error = 0;
+
+	os_rwlock_wrlock(&vinode->rwlock);
+
+	TX_BEGIN_CB(pfp->pop, cb_queue, pfp) {
+		if (!(cred->caps & (1 << PMEMFILE_CAP_CHOWN))) {
```
GBuella: These ten lines can be moved, to be outside the transaction. Also, probably there would be no need to split the 'if (owner != -1 && owner != cred->fsuid)' line into two lines, if it had one fewer tab in front of it.

src/libpmemfile-core/file.c
```diff
@@ -1674,3 +1674,171 @@ pmemfile_truncate(PMEMfilepool *pfp, const char *path, off_t length)
...
+{
+	struct pmemfile_inode *inode = vinode->inode;
+	int error = 0;
+
+	if (owner == (uid_t)-1 && group == (gid_t)-1)
+		return 0;
+
+	os_rwlock_wrlock(&vinode->rwlock);
+
+	if (!(cred->caps & (1 << PMEMFILE_CAP_CHOWN))) {
```
GBuella: Well, not that it matters a lot, but this section of code doesn't seem to use the vinode either...
So why lock the vinode before it?

GBuella: All these 'error = EXX; goto end;' pairs can be replaced with 'return EXX' statements.
The 'end' label can be removed, and locking can be simply around the transaction.
Simpler code - better code, yupee

src/libpmemfile-core/file.c
```diff
@@ -1674,3 +1674,171 @@ pmemfile_truncate(PMEMfilepool *pfp, const char *path, off_t length)
...
+
+		if (group != (gid_t)-1 && group != inode->gid) {
+			if (group != cred->fsgid && !gid_in_list(cred, group)) {
+				error = EPERM;
+				goto end;
+			}
+		}
+	}
+
+	COMPILE_ERROR_ON(offsetof(struct pmemfile_inode, gid) !=
```
GBuella: But this assertion might actually better be right next to the pmemobj_tx_add_range_direct call, to make it even more obvious why it is needed. It doesn't do anything runtime anyways.

src/libpmemfile-core/file.c
marcinslusarz: Unfortunately this code uses inode->uid/gid, so I can't move it above the lock.

src/libpmemfile-core/file.c
marcinslusarz: Done.

src/libpmemfile-core/file.c
GBuella: ooooooh, I see now, sorry

PR #66, opened 2017-03-27
Coverity fixes

marcinslusarz:
This change is [Reviewable](https://reviewable.io/reviews/marcinslusarz/pmemfile/66)

### DIFF
--- a/src/libpmemfile/preload.c
+++ b/src/libpmemfile/preload.c
@@ -90,11 +90,18 @@
 #define SYS_pwritev2 328
 #endif
 
+static pf_noreturn void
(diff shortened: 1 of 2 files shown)


src/libpmemfile/preload.c
```diff
@@ -90,11 +90,18 @@
 #define SYS_pwritev2 328
 #endif
 
+static pf_noreturn void
```
GBuella: Seriously, what is the problem with just calling it 'static void' ?

GBuella: Oh, I see, it should have the noreturn attribute to help the analyzer.

PR #73, opened 2017-03-29
posix: fix block_list_remove

GBuella:
The relocation of a block was not represented in
the tree of blocks, leaving the tree in an inconsistent
state.

---
This change is [Reviewable](https://reviewable.io/reviews/marcinslusarz/pmemfile/73)

### DIFF
--- a/src/libpmemfile-posix/block_array.c
+++ b/src/libpmemfile-posix/block_array.c
@@ -30,6 +30,7 @@
  * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
+#include "ctree.h"
 #include "layout.h"
 #include "inode.h"
 #include "internal.h"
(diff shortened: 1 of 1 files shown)


src/libpmemfile-posix/block_array.c
```diff
@@ -256,6 +257,9 @@ block_list_remove(struct pmemfile_vinode *vinode,
 		if (vinode->first_block == moving_block)
 			vinode->first_block = block;
 		relocate_block(block, moving_block);
+		ctree_remove_unlocked(vinode->blocks, block->offset, 1);
+		ctree_insert_unlocked(vinode->blocks, block->offset,
+		    (uint64_t)block);
```
marcinslusarz: It would be easier to understand if ctree_remove would be moved before relocate_block and use moving_block->offset instead of block->offset. Other than that LGTM.

GBuella: Done

PR #81, opened 2017-03-31
posix: add pmemfile_posix_fallocate interface

GBuella:
This change is [Reviewable](https://reviewable.io/reviews/marcinslusarz/pmemfile/81)

### DIFF
--- a/include/libpmemfile-posix.h
+++ b/include/libpmemfile-posix.h
@@ -311,6 +311,9 @@ int pmemfile_ftruncate(PMEMfilepool *, PMEMfile *file, off_t length);
 int pmemfile_fallocate(PMEMfilepool *, PMEMfile *file, int mode,
 		off_t offset, off_t length);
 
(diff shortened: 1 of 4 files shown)


include/libpmemfile-posix.h
```diff
@@ -311,6 +311,13 @@ int pmemfile_ftruncate(PMEMfilepool *, PMEMfile *file, off_t length);
 int pmemfile_fallocate(PMEMfilepool *, PMEMfile *file, int mode,
 		off_t offset, off_t length);
 
+static inline int
```
marcinslusarz: Please move the definition to C file. If we'll need to change something in the implementation (like add or remove some checks) it will not be possible with static inline impl.

GBuella: Done

marcinslusarz:
You need to add this function to the map file. Just to be sure it works, please add one call from tests.

PR #83, opened 2017-03-31
posix: implement O_PATH open flag

marcinslusarz:
Note: It's important to test all functions that accept PMEMfile* as one of its arguments, because mishandled O_PATH is potential security issue (see the change in "open_file").

---
This change is [Reviewable](https://reviewable.io/reviews/marcinslusarz/pmemfile/83)


GBuella:
OK, could we have this note in ``dirs.cpp`` as well?

marcinslusarz:
Done.

PR #90, opened 2017-04-07
preload: fix fchmodat

marcinslusarz:
The only flag specified in fchmodat manpage is documented as
not implemented and Linux kernel interface does not have it,
so libc doesn't fill this argument and we get garbage as flags.
Just remove it.

Ref: marcinslusarz/pmemfile#89

---
This change is [Reviewable](https://reviewable.io/reviews/marcinslusarz/pmemfile/90)

### DIFF
--- a/src/libpmemfile/preload.c
+++ b/src/libpmemfile/preload.c
@@ -634,7 +634,7 @@ static long hook_symlinkat(const char *target,
 				struct fd_desc at, const char *linkpath);
 static long hook_fchmod(long fd, mode_t mode);
 static long hook_fchmodat(struct fd_desc at, const char *path,
-				mode_t mode, int flags);
+				mode_t mode);
 static long hook_fchown(long fd, uid_t owner, gid_t group);
 static long hook_fchownat(struct fd_desc at, const char *path,
 				uid_t owner, gid_t group, int flags);
@@ -830,14 +830,14 @@ dispatch_syscall(long syscall_number,
 
 	if (syscall_number == SYS_chmod)
 		return hook_fchmodat(cwd_desc(), (const char *)arg0,
-					(mode_t)arg1, 0);
+					(mode_t)arg1);
 
 	if (syscall_number == SYS_fchmod)
 		return hook_fchmod(arg0, (mode_t)arg1);
 
 	if (syscall_number == SYS_fchmodat)
 		return hook_fchmodat(fetch_fd(arg0), (const char *)arg1,
-					(mode_t)arg2, (int)arg3);
+					(mode_t)arg2);
 
 	if (syscall_number == SYS_chown)
 		return hook_fchownat(cwd_desc(), (const char *)arg0,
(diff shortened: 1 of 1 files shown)


GBuella:
Review status: 0 of 5 files reviewed at latest revision, 1 unresolved discussion.

---

*[include/libpmemfile-posix.h, line 255 at r1](https://reviewable.io:443/reviews/marcinslusarz/pmemfile/90#-Kh7DgOH_lWOxUta_Svz:-Kh7DgOH_lWOxUta_Sw-:b-7kmfym) ([raw file](https://github.com/marcinslusarz/pmemfile/blob/2fc7d53b29772ca5f69ab8a20b2b81b2cc1467df/include/libpmemfile-posix.h#L255)):*
> ```objc
> int pmemfile_fchmod(PMEMfilepool *, PMEMfile *, mode_t mode);
> int pmemfile_fchmodat(PMEMfilepool *, PMEMfile *dir, const char *pathname,
> 	mode_t mode, int flags);
> ```

I think it is valuable to keep the interface consistent with libc, ergo keeping the flag argument in pmemfile_chmodat, and ignoring it in the implementation.
But it yes, it should be removed from hook_chmodat in libpmemfile.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/marcinslusarz/pmemfile/90)*

GBuella:
Review status: 0 of 5 files reviewed at latest revision, 1 unresolved discussion.

---

*[include/libpmemfile-posix.h, line 255 at r1](https://reviewable.io:443/reviews/marcinslusarz/pmemfile/90#-Kh7DgOH_lWOxUta_Svz:-Kh7F5BkALHdIlXI9JNV:bbakviv) ([raw file](https://github.com/marcinslusarz/pmemfile/blob/2fc7d53b29772ca5f69ab8a20b2b81b2cc1467df/include/libpmemfile-posix.h#L255)):*

I think it is valuable to keep the interface consistent with libc, ergo keeping the flag argument in pmemfile_chmodat, and ignoring it in the implementation.
But it yes, it should be removed from hook_chmodat in libpmemfile.
</blockquote></details>

In fact, I rather argue for keeping the current implementation as well.
We can follow POSIX in this regard: http://pubs.opengroup.org/onlinepubs/9699919799/functions/chmod.html
After all, the library is called libpmemfile-posix.
The flag argument should only be removed from hook_fchmodat, and it should just pass zero as flags to pmemfile_fchmodat.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/marcinslusarz/pmemfile/90)*

marcinslusarz:
OK.

GBuella:
:lgtm:

---

Reviewed 2 of 5 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/marcinslusarz/pmemfile/90#-:-Kh7e5TwqelVncQBXInj:bnfp4nl)*

PR #116, opened 2017-04-25
tests: replace pthreads with std::thread

marcinslusarz:
This change is [Reviewable](https://reviewable.io/reviews/pmem/pmemfile/116)

### DIFF
--- a/tests/posix/mt/mt.cpp
+++ b/tests/posix/mt/mt.cpp
@@ -33,8 +33,8 @@
 /*
  * mt.cpp -- multithreaded test for pmemfile_*
  */
-#include <sched.h>
-#include <unistd.h>
+#include <list>
+#include <thread>
 
 #include "pmemfile_test.hpp"
 
@@ -44,13 +44,12 @@ static PMEMfilepool *global_pfp;
 
 class mt : public pmemfile_test {
 public:
-	long ncpus;
-	pthread_t *threads;
+	unsigned ncpus;
+	std::list<std::thread> threads;
 
 	mt() : pmemfile_test(256 << 20)
 	{
-		ncpus = sysconf(_SC_NPROCESSORS_ONLN);
-		threads = new pthread_t[(size_t)ncpus * 2];
+		ncpus = std::thread::hardware_concurrency();
 	}
 
 	void
@@ -62,16 +61,12 @@ public:
 
 	~mt()
 	{
-		delete[] threads;
 	}
 };
 
-static void *
-open_close_worker(void *arg)
+static void
+open_close_worker(const char *path)
 {
-	const char *path = (const char *)arg;
-	sched_yield();
-
 	for (int i = 0; i < ops; ++i) {
 		PMEMfile *f1 = pmemfile_open(global_pfp, path, 0);
 		if (f1)
@@ -83,16 +78,11 @@ open_close_worker(void *arg)
 			}
 		}
 	}
-
-	return NULL;
 }
 
-static void *
-create_close_unlink_worker(void *arg)
+static void
+create_close_unlink_worker(const char *path)
 {
-	const char *path = (const char *)arg;
-	sched_yield();
-
 	for (int i = 0; i < ops; ++i) {
 		PMEMfile *f1 =
 			pmemfile_open(global_pfp, path, PMEMFILE_O_CREAT, 0644);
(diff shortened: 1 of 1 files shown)


krzycz:
:lgtm:

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/116#-:-Ki_YJWcDH44WDBtFINN:bnfp4nl)*

tests/posix/mt/mt.cpp
```diff
@@ -44,13 +44,12 @@ static PMEMfilepool *global_pfp;
 
 class mt : public pmemfile_test {
 public:
-	long ncpus;
-	pthread_t *threads;
+	unsigned ncpus;
+	std::list<std::thread *> threads;
```
GBuella: Why pointers?

tests/posix/mt/mt.cpp
```diff
@@ -187,18 +161,14 @@ TEST_F(mt, pread)
...
-				     (void *)file);
-		ASSERT_EQ(ret, 0) << strerror(errno);
-	}
+	for (unsigned j = 0; j < ncpus; ++j)
+		threads.push_back(new std::thread(pread_worker, file));
 
-	while (i > 0) {
-		ret = pthread_join(threads[--i], NULL);
-		ASSERT_EQ(ret, 0) << strerror(errno);
+	while (!threads.empty()) {
```
GBuella: or:

for (auto& t : threads)
  t.join();

tests/posix/mt/mt.cpp
tomaszkapela: I agree with @gbuella, no reason for pointers here. One thing that needs to be mentioned is that std::thread is non-copyable, so remember to either emplace or push r-value references, so they get moved properly.

marcinslusarz: Because I like pointers :P. Fixed.

tests/posix/mt/mt.cpp
marcinslusarz: Done.

tests/posix/mt/mt.cpp
```diff
@@ -44,13 +44,12 @@ static PMEMfilepool *global_pfp;
 
 class mt : public pmemfile_test {
 public:
-	long ncpus;
-	pthread_t *threads;
+	unsigned ncpus;
+	std::list<std::thread> threads;
```
tomaszkapela: Just one note from me, remember that std::thread is non-copyable so eplace or push rvalue references so they get moved.

GBuella: Yeah, to make the code more clear, it could be `emplace_back` instead of `push_back`

tests/posix/mt/mt.cpp
```diff
@@ -106,47 +96,29 @@ create_close_unlink_worker(void *arg)
...
-	for (int j = 0; j < ncpus / 2; ++j) {
-		ret = pthread_create(&threads[i++], NULL, open_close_worker,
-				     (void *)"/bbb");
-		ASSERT_EQ(ret, 0) << strerror(errno);
-		ret = pthread_create(&threads[i++], NULL,
-				     create_close_unlink_worker,
-				     (void *)"/bbb");
-		ASSERT_EQ(ret, 0) << strerror(errno);
+	for (unsigned j = 0; j < ncpus / 2; ++j) {
+		threads.push_back(std::thread(open_close_worker, "/bbb"));
```
GBuella: Simpler version:
`threads.emplace_back(open_close_worker, "/bbb");`

tests/posix/mt/mt.cpp
```diff
@@ -106,47 +96,29 @@ create_close_unlink_worker(void *arg)
...
-
-	for (int j = 0; j < ncpus / 2; ++j) {
-		ret = pthread_create(&threads[i++], NULL, open_close_worker,
-				     (void *)"/aaa");
-		ASSERT_EQ(ret, 0) << strerror(errno);
-		ret = pthread_create(&threads[i++], NULL,
-				     create_close_unlink_worker,
-				     (void *)"/aaa");
-		ASSERT_EQ(ret, 0) << strerror(errno);
+	for (unsigned j = 0; j < ncpus / 2; ++j) {
```
GBuella: Are we sure this should be limited by number of cpus?
There can be race issues with multiple threads on a single core as well, as I recall.

tests/posix/mt/mt.cpp
marcinslusarz: Done.

tests/posix/mt/mt.cpp
marcinslusarz: Done.

tests/posix/mt/mt.cpp
marcinslusarz: Take a closer look. It creates twice as many threads as there are cpus.

GBuella: Twice as many?
`j < ncpus / 2`
For example, if `ncpus == 1` this seems to result in:
`for (unsigned j = 0; j < 0; ++j)`

marcinslusarz: Yes. ncpus==1 is the corner case. It was like that before, so i'll make a separate change.

PR #135, opened 2017-05-11
posix: add proper rename implementation

marcinslusarz:
This change is [Reviewable](https://reviewable.io/reviews/pmem/pmemfile/135)


GBuella:
Review status: 0 of 15 files reviewed at latest revision, 1 unresolved discussion.

---

*[tests/posix/posix-helpers.cmake, line 92 at r1](https://reviewable.io:443/reviews/pmem/pmemfile/135#-KjsevwNX8nGQ7aQOG58:-KjsevwOFuzzcepoS91B:b-tlqi44) ([raw file](https://github.com/pmem/pmemfile/blob/eedf9b1ae22fd74623747dc97df947b61b0db1dd/tests/posix/posix-helpers.cmake#L92)):*
> ```CMake
> 
> 	if ($ENV{DUMP_STDOUT})
> 		file(READ ${BIN_DIR}/${TRACER}.out OUT)
> ```

Why are stdout and stderr directed into files in the first place? I forgot.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/135)*

GBuella:
Review status: 0 of 15 files reviewed at latest revision, 2 unresolved discussions.

---

*[src/libpmemfile-posix/file.c, line 1036 at r1](https://reviewable.io:443/reviews/pmem/pmemfile/135#-Kjsiiy0cH5l0g8yQ0F_:-Kjsiiy0cH5l0g8yQ0Fa:b-d9g06j) ([raw file](https://github.com/pmem/pmemfile/blob/eedf9b1ae22fd74623747dc97df947b61b0db1dd/src/libpmemfile-posix/file.c#L1036)):*
> ```C
> 	} TX_END
> 
> 	if (!error && src->vinode != dst->vinode) {
> ```

How can the vinode pointers are the same?
Doesn't that mean it is the same file with two different hardinks? I recall, an inode can't have two different vinodes. Shouldn't this condition be checked for much eariler?

Perhaps I'll understand after an hour of reading, but it is really not obvious at first sight.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/135)*

marcinslusarz:
*[src/libpmemfile-posix/file.c, line 1036 at r1](https://reviewable.io:443/reviews/pmem/pmemfile/135#-Kjsiiy0cH5l0g8yQ0F_:-KjsnyVYD0fFLBmB2MGb:bm5zu3y) ([raw file](https://github.com/pmem/pmemfile/blob/eedf9b1ae22fd74623747dc97df947b61b0db1dd/src/libpmemfile-posix/file.c#L1036)):*

How can the vinode pointers are the same?
Doesn't that mean it is the same file with two different hardinks? I recall, an inode can't have two different vinodes. Shouldn't this condition be checked for much eariler?

Perhaps I'll understand after an hour of reading, but it is really not obvious at first sight.
</blockquote></details>

pmemfile_path_info.vinode is a parent directory.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/135)*

marcinslusarz:
*[tests/posix/posix-helpers.cmake, line 92 at r1](https://reviewable.io:443/reviews/pmem/pmemfile/135#-KjsevwNX8nGQ7aQOG58:-KjsobqHICBb8NTeWNqY:bfjyvxy) ([raw file](https://github.com/pmem/pmemfile/blob/eedf9b1ae22fd74623747dc97df947b61b0db1dd/tests/posix/posix-helpers.cmake#L92)):*

Why are stdout and stderr directed into files in the first place? I forgot.
</blockquote></details>

I can't check now, but I think sometimes they are used for matching.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/135)*

src/libpmemfile-posix/HACKING.txt
```diff
@@ -0,0 +1,39 @@
...
+- When accessing vinode or inode you must hold a reference on vinode.
+- Unless there's a good reason to not do this, refing should be done outside of
+  pmemobj transaction.
+- It's safe to ref in the transaction if there's nothing after ref that can
+  abort.
+  If transaction can fail after ref, you need to handle that and unref it.
+  Also in this case you have to deal with inode_ref sometimes taking a reference
+  on its parent (if it's a directory).
+  This is very fragile and tricky, so refing inside of transaction is highly
+  discouraged.
```
marcinslusarz: FYI, the next PR will remove refing from inside of transactions, making this long point obsolete.

src/libpmemfile-posix/file.c
```diff
@@ -52,6 +53,23 @@
 #include "pool.h"
 #include "util.h"
 
+static int
+vinode_cmp(const void *v1, const void *v2)
+{
+	return (int)((intptr_t)*(void **)v1 - (intptr_t)*(void **)v2);
```
GBuella: Might wanna change this line.
I tried this a little C program, and it printed `0`:
```
int qqq(intptr_t a, intptr_t b)
{
        return (int)(a - b);
}

int main()
{
  intptr_t a = 5;
  intptr_t b = 5 + (((intptr_t)1) << 32);

  printf("qqq: %d\n", qqq(a, b));

}
```

src/libpmemfile-posix/file.c
```diff
@@ -1167,6 +1167,15 @@ _pmemfile_renameat2(PMEMfilepool *pfp,
 		goto end_unlock;
 	}
 
+	/*
+	 * From "rename" manpage:
+	 * "If oldpath and newpath are existing hard links referring to
+	 * the same file, then rename() does nothing, and returns a success
+	 * status."
+	 */
+	if (dst_vinode == src_vinode)
```
GBuella: Not that it matters a lot, but could this be checked earlier? While having the readlocks for their parents?

GBuella:
Review status: 0 of 15 files reviewed at latest revision, 2 unresolved discussions.

---

*[src/libpmemfile-posix/inode.c, line 710 at r2](https://reviewable.io:443/reviews/pmem/pmemfile/135#-KkAkarXzQQ82lW8OmIe:-KkAkarXzQQ82lW8OmIf:b-vjxjmq) ([raw file](https://github.com/pmem/pmemfile/blob/959f85a074ee3b6d4d5b886d2c47797c250f98bf/src/libpmemfile-posix/inode.c#L710)):*
> ```C
> vinode_cmp(const void *v1, const void *v2)
> {
> 	uintptr_t v1num = (uintptr_t)*(void **)v1;
> ```

`uintptr_t v1num = (uintptr_t)v1;`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/135)*

src/libpmemfile-posix/file.c
GBuella: OK, I see it is fixed in a later commit.

src/libpmemfile-posix/inode.c
```diff
@@ -626,6 +626,119 @@ inode_free(PMEMfilepool *pfp, TOID(struct pmemfile_inode) tinode)
...
+			return true;
+	return false;
+}
+
+/*
+ * vinode_wrlock4 -- take up to 4 WRITE locks on specified inodes in always
+ * the same order
+ */
+void
+vinode_wrlock4(struct pmemfile_vinode *v[], size_t *N,
```
GBuella: Can't we just null terminated the list?
That would remove the `N` argument from here, and from `vinode_unlockN`.

src/libpmemfile-posix/file.c
```diff
@@ -1192,67 +1326,23 @@ _pmemfile_renameat2(PMEMfilepool *pfp,
...
+	 * lock_parents_and_children can race with another thread messing with
+	 * source or destination directory. Loop as long as race occurs.
+	 */
+	do {
+		error = lock_parents_and_children(pfp, &src, &src_info, &dst,
+				&dst_info, vinodes, &vinodes_num);
+	} while (error == 1);
+
+	if (error < 0) {
+		error = -error;
```
GBuella: Well, if the code for restarting would be negative, and the regular errno numbers would be returned as positive, this would be a tiny bit more simple here.

src/libpmemfile-posix/file.c
```diff
@@ -997,6 +997,143 @@ pmemfile_unlink(PMEMfilepool *pfp, const char *pathname)
...
+
+	if (dst_info->vinode) {
+		vinode_unref(pfp, dst_info->vinode);
+		dst_info->vinode = NULL;
+	}
+
+	src_info->dirent = NULL;
+	dst_info->dirent = NULL;
+
+	return 1;
```
GBuella: Or perhaps it could tailcall itself, or perhaps `goto again`, `goto restart`?

src/libpmemfile-posix/file.c
```diff
@@ -997,6 +997,143 @@ pmemfile_unlink(PMEMfilepool *pfp, const char *pathname)
...
+	}
+
+	/* find destination file (it may not exist) */
+	dst_info->dirent = vinode_lookup_dirent_by_name_locked(pfp, dst->vinode,
+					dst->remaining, dst_namelen);
+	if (dst_info->dirent) {
+		/* if file exists get the vinode for it */
+		dst_info->vinode = inode_ref(pfp, dst_info->dirent->inode,
+				dst->vinode, NULL, dst->remaining, dst_namelen);
+		if (!dst_info->vinode) {
```
GBuella: Just asking, the `inode_ref` thing can only fail, if an underlying `malloc` fails -- is that right?

marcinslusarz: ... or when inode version does not match what we expect.

tests/posix/mt/mt.cmake
```diff
@@ -33,6 +33,7 @@ include(${SRC_DIR}/../posix-helpers.cmake)
 
 setup()
 
-execute(${TEST_EXECUTABLE} ${ops})
+execute(${TEST_EXECUTABLE} ${ops} --gtest_filter=mt.open_close_create_unlink)
```
GBuella: `tests: run multi-threaded testcases as separate processes`
Why?

src/libpmemfile-posix/file.c
```diff
@@ -986,21 +986,21 @@ vinode_exchange(PMEMfilepool *pfp,
 		 * If both are regular files or have the same parent, then
 		 * we don't have to do anything.
 		 */
-		if ((src_is_dir || dst_is_dir) && src_parent != dst_parent) {
+		if ((src_is_dir || dst_is_dir) && src->vinode != dst->vinode) {
```
GBuella: BTW, since vinode in path_info always points some parent directory:
```
/* path resolution information */
struct pmemfile_path_info {
        /*
         * Vinode of the last reachable component in the path, except for
         * the last part.
         */
        struct pmemfile_vinode *vinode;
```
Could it be called `parent_vinode`, or something similar?

GBuella:
One of the commit messages says:
"posix: fix crash in exchange when the first path is a parent of the second path"
But looking at the code, it should be something like:
"posix: fix crash in exchange when any the first and second path is a parent of the other"

src/libpmemfile-posix/file.c
```diff
@@ -1210,7 +1210,8 @@ _pmemfile_renameat2(PMEMfilepool *pfp,
 		goto end;
 	}
 
-	os_rwlock_wrlock(&pfp->super_rwlock);
+	if (src.vinode != dst.vinode)
+		os_rwlock_wrlock(&pfp->super_rwlock);
```
GBuella: Right.
Also, could you add the explanation here in comment as well?
```
rename(/a/b -> /1/2/3/4/5) locks a, b, 4, 5
rename(/1/2/ -> /a/b/c/d/e) locks 1, 2, d, e
The first one suceeds, now there is:
a/ and /1/2/3/4/5/c/d/e
The second one now sees, that 2 is a parent of e

And the super lock is needed to be able to look at the parents in both threads.
```

tests/posix/mt/mt.cmake
marcinslusarz: To speed up execution under Valgrind.

src/libpmemfile-posix/dir.c
```diff
@@ -1242,23 +1226,46 @@ _pmemfile_rmdirat(PMEMfilepool *pfp, struct pmemfile_vinode *dir,
 		goto vparent_end;
 	}
 
+	dirent = NULL;
```
marcinslusarz: To prevent future me (or any developer) from accidentally using it after unlock.

tests/posix/mt/mt.cmake
GBuella: `ctest -j n` doesn't help run the tests faster?
I think, if this is needed, than these should definitely be added with two `add_test` commands.

src/libpmemfile-posix/file.c
marcinslusarz: I think we need to verify file permissions first.

tests/posix/mt/mt.cmake
marcinslusarz: Good idea.

marcinslusarz:
@GBuella  WRT commit message for "posix: fix crash in exchange when the first path is a parent of the second path" - the scenario you are talking about was already handled.

GBuella:
"the scenario you are talking about was already handled." yes, I see now, only the test added in the commit was for the other scenario.

src/libpmemfile-posix/inode.c
marcinslusarz: We could, but then it would mean we have to pass an array of at least size 5, which would look a bit weird...

I realized memset is not needed - I'll remove it.

GBuella: It might look weird to you, but it removes a variable. Of which, there are already too many for most of us to follow.
The compiler can check it for you.
```
int func(int x[static 5])
{
        return x[2] + x[3];
}

int v[4];
func(vv);

test.c:67:2: warning: array argument is too small; contains 4 elements, callee requires at least 5 [-Warray-bounds]
        func(vv);
```
Anyways, couldn't this be a function with variadic arguments, with a null terminated argument list? When supplied with the right GNU attributes, the compiler can also check for the existence of the null terminator.

tests/posix/dirs/dirs.cpp
```diff
@@ -214,7 +216,7 @@ TEST_F(dirs, lots_of_files)
 	ASSERT_TRUE(test_empty_dir(pfp, "/"));
 	memset(buf, 0xff, sizeof(buf));
 
-	for (size_t i = 0; i < 100; ++i) {
+	for (size_t i = 0; i < ops; ++i) {
 		sprintf(buf, "/file%04lu", i);
```
marcinslusarz: Done.

tests/posix/dirs/dirs.cpp
```diff
@@ -231,25 +233,26 @@ TEST_F(dirs, lots_of_files)
...
 
-	EXPECT_TRUE(
-		test_compare_dirs(pfp, "/", std::vector<pmemfile_ls>{
-						    {040777, 2, 32680, "."},
-						    {040777, 2, 32680, ".."},
-					    }));
+	if (ops == 100)
+		EXPECT_TRUE(test_compare_dirs(pfp, "/",
+					      std::vector<pmemfile_ls>{
+						      {040777, 2, 32680, "."},
```
marcinslusarz: Done.

src/libpmemfile-posix/dir.c
```diff
@@ -547,8 +532,6 @@ vinode_unlink_dirent(PMEMfilepool *pfp, struct pmemfile_vinode *parent,
 	 */
 	TX_SET_DIRECT(parent->inode, mtime, tm);
 
-	rwlock_tx_unlock_on_commit(&(*vinode)->rwlock);
-
 	dirent->name[0] = '\0';
```
marcinslusarz: Done.

marcinslusarz:
*[src/libpmemfile-posix/inode.c, line 710 at r2](https://reviewable.io:443/reviews/pmem/pmemfile/135#-KkAkarXzQQ82lW8OmIe:-KkGUTS-z4TSgqYTjVCV:b-wp841v) ([raw file](https://github.com/pmem/pmemfile/blob/959f85a074ee3b6d4d5b886d2c47797c250f98bf/src/libpmemfile-posix/inode.c#L710)):*

`uintptr_t v1num = (uintptr_t)v1;`
</blockquote></details>

Your version is incorrect. I want to compare addresses on vinodes, not their positions in an array.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/135)*

GBuella:
Reviewed 12 of 15 files at r1, 3 of 3 files at r2.

---

*[src/libpmemfile-posix/inode.c, line 710 at r2](https://reviewable.io:443/reviews/pmem/pmemfile/135#-KkAkarXzQQ82lW8OmIe:-KkGWL-JjxSvyVNIF36V:b4z3q81) ([raw file](https://github.com/pmem/pmemfile/blob/959f85a074ee3b6d4d5b886d2c47797c250f98bf/src/libpmemfile-posix/inode.c#L710)):*

Your version is incorrect. I want to compare addresses on vinodes, not their positions in an array.
</blockquote></details>

Oops, right!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/135)*

src/libpmemfile-posix/inode.c
marcinslusarz: Ok, I null-terminated an output array.

src/libpmemfile-posix/file.c
marcinslusarz: IIRC some time ago we discussed the plan to change how errno is handled (without using errno variable and passing negated errno codes) and some functions already use this new convention. I just used it here.

src/libpmemfile-posix/file.c
marcinslusarz: Recursive functions and back jumps make the code hard to analyze... I'd prefer to not use them unless they are really necessary.

src/libpmemfile-posix/file.c
```diff
@@ -1252,14 +1247,32 @@ vinode_rename(PMEMfilepool *pfp,
...
 
-		if (vinode_is_dir(src_vinode) && src_parent != dst_parent)
-			vinode_update_parent(pfp, src_vinode, src_parent,
-					dst_parent);
+			/*
+			 * From "stat" man page:
+			 * "st_mtime of a directory is changed by the creation
+			 * or deletion of files in that directory."
+			 */
+			TX_SET_DIRECT(src_parent->inode, mtime, t);
```
marcinslusarz: No. mtime is modified in vinode_add_dirent and vinode_unlink_file called when src_parent != dst_parent.

src/libpmemfile-posix/file.c
marcinslusarz: I'll do that in the next PR.

tests/posix/mt/mt.cmake
marcinslusarz: Done.

src/libpmemfile-posix/file.c
marcinslusarz: Done.

GBuella:
Reviewed 1 of 1 files at r3, 4 of 5 files at r5, 2 of 2 files at r6, 1 of 1 files at r7.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/135)*

PR #141, opened 2017-05-18
tests: fix TESTS_USE_FORCED_PMEM=ON

marcinslusarz:
"if (${VAR})" evaluates to "if (ON)" when VAR=ON. So now cmake evaluates
"if (ON)" which is not true, because it tries to evaluate variable
"ON" which is not set.

"if (VAR)" works both when VAR=1 or VAR=ON.

---
This change is [Reviewable](https://reviewable.io/reviews/pmem/pmemfile/141)

### DIFF
--- a/tests/posix/posix-helpers.cmake
+++ b/tests/posix/posix-helpers.cmake
@@ -40,12 +40,12 @@ function(cleanup)
 endfunction()
 
 function(execute name)
-	if(${TESTS_USE_FORCED_PMEM})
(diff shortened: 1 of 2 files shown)


krzycz:
:lgtm:

---

Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/141#-:-KkQcHBkE7av_aOBfKl8:bnfp4nl)*

PR #152, opened 2017-05-26
Refactor compile time checks

GBuella:
This change is [Reviewable](https://reviewable.io/reviews/pmem/pmemfile/152)


marcinslusarz:
ASSERT(1) doesn't work on purpose. We had a bug in nvml, where someone used ASSERT(1) instead of ASSERT(0). Also asserting something that is always true (because compiler can see it is at compile time) is useless. COMPILE_ERROR_ON is supposed to be used for that purpose.

This is explained in the comment just above the stuff you removed.

GBuella:
Please re-read the commit message, I deliberately kept the original behaviour of `ASSERT`.

marcinslusarz:
Ok. I misunderstood your patch. Please keep the old version of COMPILE_ERROR_ON under different name and use it in ASSERT macros.

marcinslusarz:
Ha, I just recalled why __builtin_constant_p was needed - to not evaluate the condition more than 1 time.

marcinslusarz:
Reviewed 3 of 7 files at r1, 3 of 3 files at r2, 2 of 2 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/152)*

marcinslusarz:
:lgtm:

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/152#-:-Kl4UUOWHnBwuA44qZU1:bnfp4nl)*

GBuella:
"Ha, I just recalled why __builtin_constant_p was needed - to not evaluate the condition more than 1 time." I still don't understand that. Could you elaborate?

marcinslusarz:
With unconditional COMPILE_ERROR_ON cnd would be evaluated both in COMPILER_ERROR_ON and in ASSERT_rt.

GBuella:
It is evaluated in both places, wether that 'if' is there or not.

PR #248, opened 2017-07-12
preload: fix forward declarations in preload.h

GBuella:
The only reason these happened to work before, was that the
libpmemfile-posix.h header was always included before
the preload.h header.

---
This change is [Reviewable](https://reviewable.io/reviews/pmem/pmemfile/248)

### DIFF
--- a/src/libpmemfile/preload.h
+++ b/src/libpmemfile/preload.h
@@ -39,8 +39,8 @@
 #include <pthread.h>
 #include <linux/limits.h>
 
-struct PMEMfilepool;
-struct PMEMfile;
+struct pmemfilepool;
+struct pmemfile_file;
 
 struct pool_description {
 	/*
@@ -79,7 +79,7 @@ struct pool_description {
 	 * If this is NULL, the mount point was not used by the application
 	 * before in this process. Should be initialized on first use.
 	 */
-	PMEMfilepool *pool;
+	struct pmemfilepool *pool;
 
 	/* Data about the root directory inside the pmemfile pool */
 	struct stat pmem_stat;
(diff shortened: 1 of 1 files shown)


krzycz:
:lgtm:

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/248#-:-KosMmCzLlvkhmvpYoxI:bnfp4nl)*

marcinslusarz:
Why not include libpmemfile-posix.h instead?

GBuella:
Why not include libpmemfile-posix.h instead?
Usually I prefer this approach -> fewer dependencies -> less complexity.
Would you rather libpmemfile-posix.h to be included?

marcinslusarz:
My approach to forward declarations is "don't use them unless there's no other way" :). That's because such header file is incomplete - if you want want to use you have to include other header file too.

So I'd prefer if you would include libpmemfile-posix.h.

GBuella:
Two errors in that reasoning:
1) there is no need to include other headers to use this one
2) libpmemfile-posix.h also contains only the same forward declaration of a file pointer, thus turning your weapon against yourself... -- would you like me to include `src/libpmemfile-posix/file.h` ?

marcinslusarz:
1. Right
2. "no other way" (without exposing internal structure definition) rule applies here

marcinslusarz:
:lgtm:

---

Reviewed 1 of 1 files at r1.
Review status: :shipit: all files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/248#-:-KpJwv_8oif74NOJnlaZ:bnfp4nl)*

PR #262, opened 2017-07-19
posix: fix errno for lseek to CUR and END

szychows:
EOVERFLOW errno is only relevant with 32-bit off_t

---
This change is [Reviewable](https://reviewable.io/reviews/pmem/pmemfile/262)

### DIFF
--- a/src/libpmemfile-posix/lseek.c
+++ b/src/libpmemfile-posix/lseek.c
@@ -195,6 +195,19 @@ pmemfile_lseek_locked(PMEMfilepool *pfp, PMEMfile *file, pmemfile_off_t offset,
 				 * From POSIX: EINVAL if
 				 * "...the resulting file offset would be
 				 * negative for a regular file..."
+				 *
+				 * POSIX manpage also mentions EOVERFLOW
+				 * "The resulting file offset would be a value
+				 * which cannot be represented correctly in an
+				 * object of type off_t."
+				 * However in existing implementations it looks
+				 * like it is only used to represent user-type
+				 * overflow - user calls lseek, when off_t is
+				 * 32-bit, but internal kernel type is 64-bit,
+				 * and returned value cannot be represented
+				 * EOVERFLOW is returned.
+				 * With 64-bit off_t type EINVAL is returned in
+				 * case of overflow.
 				 */
 				new_errno = EINVAL;
 			}
@@ -202,18 +215,8 @@ pmemfile_lseek_locked(PMEMfilepool *pfp, PMEMfile *file, pmemfile_off_t offset,
 		case PMEMFILE_SEEK_CUR:
 			ret = (pmemfile_off_t)file->offset + offset;
 			if (ret < 0) {
-				if (offset < 0) {
-					new_errno = EINVAL;
-				} else {
-					/*
-					 * From POSIX: EOVERFLOW if
-					 * "...The resulting file offset would
-					 * be a value which cannot be
-					 * represented correctly in an object
-					 * of type off_t..."
-					 */
-					new_errno = EOVERFLOW;
-				}
+				/* Error as in SEEK_SET */
+				new_errno = EINVAL;
 			}
 			break;
 		case PMEMFILE_SEEK_END:
@@ -221,11 +224,8 @@ pmemfile_lseek_locked(PMEMfilepool *pfp, PMEMfile *file, pmemfile_off_t offset,
 			ret = (pmemfile_off_t)inode->size + offset;
 			os_rwlock_unlock(&vinode->rwlock);
 			if (ret < 0) {
-				/* Errors as in SEEK_CUR */
-				if (offset < 0)
-					new_errno = EINVAL;
-				else
-					new_errno = EOVERFLOW;
+				/* Error as in SEEK_SET */
+				new_errno = EINVAL;
 			}
 			break;
 		case PMEMFILE_SEEK_DATA:
--- a/tests/posix/rw/rw.cpp
+++ b/tests/posix/rw/rw.cpp
@@ -237,13 +237,13 @@ TEST_F(rw, basic)
 	ASSERT_EQ(errno, EINVAL);
 	errno = 0;
 	ASSERT_EQ(pmemfile_lseek(pfp, f, INT64_MAX, PMEMFILE_SEEK_CUR), -1);
-	ASSERT_EQ(errno, EOVERFLOW);
+	ASSERT_EQ(errno, EINVAL);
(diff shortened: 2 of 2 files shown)


marcinslusarz:
:lgtm:

---

Reviewed 2 of 2 files at r1.
Review status: :shipit: all files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/262#-:-KpPhlETY9cVMC3R2-H_:bnfp4nl)*

GBuella:
But that is totally not what the man page says.
Is this done due looking at what lseek returns on ext4?

szychows:
pmemfile_off_t is 64 bit type (typedef to off64_t). 

EOVERFLOW applies only to 32-bit off_t in a situation, where resulting offset would be bigger than value that you can store in 32-bit variable.

libpmemfile-posix does not have separate pmemfile_lseek64 and it takes pmemfile_off_t as an argument, so I think we can assume pmemfile_lseek is an equivalent to lseek64.

GBuella:
I'm not sure what you mena, a 64 bit type can overflow just as well. It can not store a value larger than 2^63-1 , so it e.g.
if `offset ==2^63 - 1` then `lseek(fd, 4, SEEK_CUR)` should mean an overflow, since that signed 64 bit int can not represent the number 2^63 + 3

Nonetheless, I checked stuff with a dummy program, and it really seems to return EINVAL.

What I noticed, is ext4 just returns EINVAL for anything that points over 16 terabytes, which I guess is the maximum file size on my machine.
When I test things in `/dev/shm`, anything up to `INT64_MAX` is allowed, but after an `lseek(fd, 1, SEEK_CUR)` returns EINVAL, instead of EOVERFLOW.
So the man page looks like it follows POSIX, but the implementation does not always...

szychows:
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/fs/read_write.c#n315

A lucky search found me this.
This comment says that it can only happen on 32-bit systems.

src/libpmemfile-posix/lseek.c
```diff
@@ -202,30 +202,17 @@ pmemfile_lseek_locked(PMEMfilepool *pfp, PMEMfile *file, pmemfile_off_t offset,
...
-					/*
-					 * From POSIX: EOVERFLOW if
-					 * "...The resulting file offset would
-					 * be a value which cannot be
-					 * represented correctly in an object
-					 * of type off_t..."
-					 */
-					new_errno = EOVERFLOW;
-				}
+				/* Error as in SEEK_SET */
```
GBuella: And please add a little comment about following some existing implementations instead of POSIX, or the lseek man page on linux, so we know why this is so.

GBuella:
Yes, well then the one who wrote that comment does not know arithmetic or something. This is pretty unequivocal:
http://man7.org/linux/man-pages/man2/lseek.2.html
"EOVERFLOW  The resulting file offset cannot be represented in an off_t."
or:
http://pubs.opengroup.org/onlinepubs/9699919799/functions/lseek.html
"EOVERFLOW  The resulting file offset would be a value which cannot be represented correctly in an object of type off_t."

I guess the routine `fn` there should return `EOVERFLOW` in such case:
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/fs/read_write.c#n299
because that is the routine which should add the two integers.

But apparently some or most fs implementors don't do that, so we can follow that, whatever.

marcinslusarz:
As I understand it EINVAL is used when overflow happens for in-kernel offset representation (which is always 64-bit) and EOVERFLOW for user type (off_t, which can be 32-bit).

lseek manpage also says:
EINVAL whence is not valid.  Or: **the resulting file offset would be negative**, or beyond the end of a seekable device.

So both EINVAL and EOVERFLOW are used for overflowing, but for different variables.

szychows:
Review status: 0 of 2 files reviewed at latest revision, 2 unresolved discussions.

---

*[src/libpmemfile-posix/lseek.c, line 203 at r1](https://reviewable.io:443/reviews/pmem/pmemfile/262#-KpPdlog-w5o7HNFuipy-r1-203:-Kpsunn8u34uE7Hz03yw:b-2px7gj) ([raw file](https://github.com/pmem/pmemfile/blob/c1a07ff6ba4e8aef5e6d5ebf43f8720caac16c0d/src/libpmemfile-posix/lseek.c#L203)):*

At the same time, perhaps we could get rid of UB in these signed integer additions.
</blockquote></details>

So what would be desired behaviour?

---

*[src/libpmemfile-posix/lseek.c, line 205 at r1](https://reviewable.io:443/reviews/pmem/pmemfile/262#-KpPdlog-w5o7HNFuipy-r1-205:-KpsunG42bKp__moRx2F:b-896fix) ([raw file](https://github.com/pmem/pmemfile/blob/c1a07ff6ba4e8aef5e6d5ebf43f8720caac16c0d/src/libpmemfile-posix/lseek.c#L205)):*

And please add a little comment about following some existing implementations instead of POSIX, or the lseek man page on linux, so we know why this is so.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/262)*

krzycz:
Reviewed 2 of 2 files at r2.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/262)*

GBuella:
:lgtm:

---

---

*[src/libpmemfile-posix/lseek.c, line 203 at r1](https://reviewable.io:443/reviews/pmem/pmemfile/262#-KpPdlog-w5o7HNFuipy-r1-203:-Kpt8QqGLy_v8pHb8DnO:b-sx1ov0) ([raw file](https://github.com/pmem/pmemfile/blob/c1a07ff6ba4e8aef5e6d5ebf43f8720caac16c0d/src/libpmemfile-posix/lseek.c#L203)):*
<details><summary><i>Previously, szychows wrote…</i></summary><blockquote>

So what would be desired behaviour?
</blockquote></details>

The desired behavior is well defined behavior.
I think if both are non-negative, cast them to the unsigned version of pmemfile_off_t, add the unsigned ones, and if the result is not representable in the original signed version, we know it is EINVAL. But we can do that in a different commit I guess.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/262#-:-KptAN221ls6_PlJKbE4:bnfp4nl)*

PR #283, opened 2017-07-28
Use libclang to generate some wrapper functions

GBuella:
This change is [Reviewable](https://reviewable.io/reviews/pmem/pmemfile/283)

PR #291, opened 2017-08-03
vfd_table and dup

GBuella:
This is just an RFC as of now.
This PR contains the commits from #283 .
Also needs more comments and whatnot.

---
This change is [Reviewable](https://reviewable.io/reviews/pmem/pmemfile/291)

---

As one of the side effects, this removes the need for a special cwd_lock being used for all syscalls that are affected by the CWD.
Thus, all syscalls such as `open`, `stat`, etc.. using paths were serialized before, but with this patch, they can run in paralell.

---

Update:
Now that #283 was merged, I rebased it on the current master. But there were a lot of conflicts with the process switching related changes, so this is still work in progress.
I'm sure there are many places where pool_acquire/pool_release are not used correctly, it will take a while clear it up and be confident about it.

### DIFF
--- a/src/libpmemfile/preload.c
+++ b/src/libpmemfile/preload.c
@@ -1830,6 +1830,12 @@ dispatch_syscall(long syscall_number,
 		return hook_bind((int)arg0, (const struct sockaddr *)arg1,
 				(socklen_t)arg2);
 
+	case SYS_dup:
+		return pmemfile_vfd_dup((int)arg0);
+
+	case SYS_dup2:
+		return pmemfile_vfd_dup2((int)arg0, (int)arg1);
+
 	default:
 		/* Did we miss something? */
 		assert(false);
--- a/src/libpmemfile/preload.c
+++ b/src/libpmemfile/preload.c
@@ -2077,6 +2077,19 @@ hook(long syscall_number,
 		return HOOKED;
 	}
 
+	if (syscall_number == SYS_fcntl &&
+	    ((int)arg1 == F_DUPFD || (int)arg1 == F_DUPFD_CLOEXEC)) {
+		/*
+		 * Other fcntl commands on pmemfile resident files are handled
+		 * via dispatch_syscall_fd_first.
+		 *
+		 * XXX: close-on-exec flag is not handled correctly yet.
+		 */
+		*syscall_return_value =
+		    pmemfile_vfd_fcntl_dup((int)arg0, (int)arg2);
+		return HOOKED;
+	}
+
 	struct syscall_early_filter_entry filter_entry;
 	filter_entry = get_early_filter_entry(syscall_number);
 
--- a/src/libpmemfile/preload.c
+++ b/src/libpmemfile/preload.c
@@ -117,10 +117,10 @@ log_write(const char *fmt, ...)
 	syscall_no_intercept(SYS_write, log_fd, buf, len);
 }
 
+static pthread_mutex_t cwd_mutex = PTHREAD_MUTEX_INITIALIZER;
+
 static struct pool_description pools[0x100];
 static int pool_count;
-
-static bool is_memfd_syscall_available;
 
 #define PMEMFILE_MAX_FD 0x8000
 
@@ -182,17 +182,6 @@ pool_release(struct pool_description *pool)
 	errno = oerrno;
 }
 
-struct pmemfile_entry {
-	struct fd_association pmemfile;
-	int ref_count;
-};
-
-/*
- * The associations between user visible fd numbers and
- * pmemfile pointers. Each pmemfile fd has it's own reference counter.
- */
-static struct pmemfile_entry fd_table[PMEMFILE_MAX_FD + 1];
-
 /*
  * A separate place to keep track of fds used to hold mount points open, in
  * the previous array. The application should not be aware of these, thus
@@ -200,91 +189,6 @@ static struct pmemfile_entry fd_table[PMEMFILE_MAX_FD + 1];
  * must be returned. The contents of this array does not change after startup.
  */
 static bool mount_point_fds[PMEMFILE_MAX_FD + 1];
-
-static bool
-is_fd_in_table(long fd)
-{
-	if (fd < 0 || fd > PMEMFILE_MAX_FD)
-		return false;
-
-	return !is_fda_null(&fd_table[fd].pmemfile);
-}
-
-static pthread_rwlock_t pmem_cwd_lock = PTHREAD_RWLOCK_INITIALIZER;
-static struct pool_description *volatile cwd_pool;
-
-static pthread_mutex_t fd_table_mutex = PTHREAD_MUTEX_INITIALIZER;
-
-static void
-fd_unref(long fd, struct fd_association *file)
-{
-	if (__sync_sub_and_fetch(&fd_table[fd].ref_count, 1) == 0) {
-		(void) syscall_no_intercept(SYS_close, fd);
-
-		struct pool_description *pool = file->pool;
-
-		pool_acquire(pool);
-
-		fd_first_pmemfile_close(file);
-
-		pool_release(pool);
-	}
-}
-
-static struct fd_association
-fd_ref(long fd)
-{
-	struct fd_association file;
-	file.file = NULL;
-	file.pool = NULL;
-
-	util_mutex_lock(&fd_table_mutex);
-
-	if (is_fd_in_table(fd)) {
-		__sync_add_and_fetch(&fd_table[fd].ref_count, 1);
-		file = fd_table[fd].pmemfile;
-	}
-
-	util_mutex_unlock(&fd_table_mutex);
-
-	return file;
-}
-
-static struct fd_desc
-cwd_desc(void)
-{
-	struct fd_desc result;
-
-	result.kernel_fd = AT_FDCWD;
-	result.pmem_fda.pool = cwd_pool;
-	result.pmem_fda.file = PMEMFILE_AT_CWD;
-
-	return result;
-}
-
-static struct fd_desc
-fd_fetch(long fd)
-{
-	struct fd_desc result;
-
-	result.kernel_fd = fd;
-
-	if ((int)fd == AT_FDCWD) {
-		result.pmem_fda.pool = cwd_pool;
-		result.pmem_fda.file = PMEMFILE_AT_CWD;
-	} else {
-		result.pmem_fda = fd_ref(fd);
-	}
-
-	return result;
-}
-
-static void
-fd_release(struct fd_desc *at)
-{
-	if (!is_fda_null(&at->pmem_fda) && at->pmem_fda.file != PMEMFILE_AT_CWD)
-		fd_unref(at->kernel_fd, &at->pmem_fda);
-}
 
 static int exit_on_ENOTSUP;
 static long check_errno(long e, long syscall_no)
@@ -298,52 +202,6 @@ static long check_errno(long e, long syscall_no)
 	}
 
 	return e;
-}
-
-#ifdef SYS_memfd_create
-
-static void
-check_memfd_syscall(void)
-{
-	long fd = syscall_no_intercept(SYS_memfd_create, "check", 0);
-	if (fd >= 0) {
-		is_memfd_syscall_available = true;
-		syscall_no_intercept(SYS_close, fd);
-	}
-}
-
-#else
-
-#define SYS_memfd_create 0
-#define check_memfd_syscall()
-
-#endif
-
-/*
- * acquire_new_fd - grab a new file descriptor from the kernel
- */
-static long
-acquire_new_fd(const char *path)
-{
-	long fd;
-
-	if (is_memfd_syscall_available) {
-		fd = syscall_no_intercept(SYS_memfd_create, path, 0);
-		/* memfd_create can fail for too long name */
-		if (fd < 0) {
-			fd = syscall_no_intercept(SYS_open, "/dev/null",
-					O_RDONLY);
-		}
-	} else {
-		fd = syscall_no_intercept(SYS_open, "/dev/null", O_RDONLY);
-	}
-
-	if (fd > PMEMFILE_MAX_FD) {
-		syscall_no_intercept(SYS_close, fd);
-		return -ENFILE;
-	}
-
-	return fd;
 }
 
 void
@@ -368,40 +226,14 @@ exit_with_msg(int ret, const char *msg)
 }
 
 static long
-hook_close(long fd)
-{
-	bool is_fd_pmem = false;
-	struct fd_association file;
-
-	util_mutex_lock(&fd_table_mutex);
-
-	if (is_fd_in_table(fd)) {
-		file = fd_table[fd].pmemfile;
-		fd_table[fd].pmemfile.file = NULL;
-		fd_table[fd].pmemfile.pool = NULL;
-
-		is_fd_pmem = true;
-	}
-
-	util_mutex_unlock(&fd_table_mutex);
-
-	if (is_fd_pmem)
-		fd_unref(fd, &file);
-	else
-		(void) syscall_no_intercept(SYS_close, fd);
-
-	return 0;
-}
-
-static long
-hook_linkat(long fd0, long arg0, long fd1, long arg1, long flags)
+hook_linkat(int fd0, long arg0, int fd1, long arg1, long flags)
 {
 	long ret;
 	struct resolved_path where_old;
 	struct resolved_path where_new;
 
-	struct fd_desc at0 = fd_fetch(fd0);
-	struct fd_desc at1 = fd_fetch(fd1);
+	struct vfd_reference at0 = pmemfile_vfd_at_ref(fd0);
+	struct vfd_reference at1 = pmemfile_vfd_at_ref(fd1);
 
 	resolve_path(at0, (const char *)arg0, &where_old, RESOLVE_LAST_SLINK);
 	resolve_path(at1, (const char *)arg1, &where_new,
@@ -411,65 +243,62 @@ hook_linkat(long fd0, long arg0, long fd1, long arg1, long flags)
 		ret = where_old.error_code;
 	} else if (where_new.error_code != 0) {
 		ret = where_new.error_code;
-	} else if (where_new.at.pmem_fda.pool != where_old.at.pmem_fda.pool) {
+	} else if (where_new.at_pool != where_old.at_pool) {
 		/* cross-pool link are not possible */
 		ret = -EXDEV;
-	} else if (is_fda_null(&where_new.at.pmem_fda)) {
+	} else if (where_new.at_pool == NULL) {
 		ret = syscall_no_intercept(SYS_linkat,
-		    where_old.at.kernel_fd, where_old.path,
-		    where_new.at.kernel_fd, where_new.path, flags);
+		    where_old.at_kernel, where_old.path,
+		    where_new.at_kernel, where_new.path, flags);
 	} else {
-		struct pool_description *pool = where_old.at.pmem_fda.pool;
+		pool_acquire(where_old.at_pool);
 
-		pool_acquire(pool);
-
-		int r = wrapper_pmemfile_linkat(pool->pool,
-			    where_old.at.pmem_fda.file, where_old.path,
-			    where_new.at.pmem_fda.file, where_new.path,
+		int r = wrapper_pmemfile_linkat(
+			    where_old.at_pool->pool,
+			    where_old.at_dir, where_old.path,
+			    where_new.at_dir, where_new.path,
 				(int)flags);
 
-		pool_release(pool);
+		pool_release(where_old.at_pool);
 
 		ret = check_errno(r, SYS_linkat);
 	}
 
-	fd_release(&at0);
-	fd_release(&at1);
+	pmemfile_vfd_unref(at0);
+	pmemfile_vfd_unref(at1);
 
 	return ret;
 }
 
 static long
-hook_unlinkat(long fd, long path_arg, long flags)
+hook_unlinkat(int fd, long path_arg, long flags)
 {
 	long ret;
 	struct resolved_path where;
 
-	struct fd_desc at = fd_fetch(fd);
+	struct vfd_reference at = pmemfile_vfd_at_ref(fd);
 
 	resolve_path(at, (const char *)path_arg,
 	    &where, NO_RESOLVE_LAST_SLINK);
 
 	if (where.error_code != 0) {
 		ret = where.error_code;
-	} else if (is_fda_null(&where.at.pmem_fda)) {
+	} else if (where.at_pool == NULL) {
 		/* Not pmemfile resident path */
 		ret = syscall_no_intercept(SYS_unlinkat,
-				where.at.kernel_fd, where.path, flags);
+				where.at_kernel, where.path, flags);
 	} else {
-		struct pool_description *pool = where.at.pmem_fda.pool;
+		pool_acquire(where.at_pool);
 
-		pool_acquire(pool);
+		int r = wrapper_pmemfile_unlinkat(where.at_pool->pool,
+			where.at_dir, where.path, (int)flags);
 
-		int r = wrapper_pmemfile_unlinkat(pool->pool,
-			where.at.pmem_fda.file, where.path, (int)flags);
-
-		pool_release(pool);
+		pool_release(where.at_pool);
 
 		ret = check_errno(r, SYS_unlinkat);
 	}
 
-	fd_release(&at);
+	pmemfile_vfd_unref(at);
 
 	return ret;
 
@@ -478,126 +307,99 @@ hook_unlinkat(long fd, long path_arg, long flags)
 static long
 hook_chdir(const char *path)
 {
+	util_mutex_lock(&cwd_mutex);
 	struct resolved_path where;
 
 	long result;
 
 	log_write("%s(\"%s\")", __func__, path);
 
-	util_rwlock_wrlock(&pmem_cwd_lock);
-
-	resolve_path(cwd_desc(), path, &where, RESOLVE_LAST_SLINK);
+	struct vfd_reference at = pmemfile_vfd_at_ref(AT_FDCWD);
+	resolve_path(at, path, &where, RESOLVE_LAST_SLINK);
 
 	if (where.error_code != 0) {
 		result = where.error_code;
-	} else if (is_fda_null(&where.at.pmem_fda)) {
-		result = syscall_no_intercept(SYS_chdir, where.path);
-		if (result == 0)
-			cwd_pool = NULL;
-	} else {
-		if (cwd_pool != where.at.pmem_fda.pool) {
-			cwd_pool = where.at.pmem_fda.pool;
-			syscall_no_intercept(SYS_chdir, cwd_pool->mount_point);
-		}
+	} else if (where.at_pool == NULL) {
+		long fd = syscall_no_intercept(SYS_openat,
+				where.at_kernel, where.path,
+				O_DIRECTORY | O_PATH | O_NOCTTY);
 
-		pool_acquire(cwd_pool);
-
-		if (pmemfile_chdir(cwd_pool->pool, where.path) == 0)
-			result = 0;
+		if (fd >= 0)
+			result = pmemfile_vfd_chdir_kernel_fd((int)fd);
 		else
+			result = fd;
+	} else {
+		pool_acquire(where.at_pool);
+
+		PMEMfile *file = pmemfile_openat(where.at_pool->pool,
+					where.at_dir, where.path,
+					O_DIRECTORY | O_PATH | O_NOCTTY);
+
+		if (file == NULL) {
 			result = -errno;
-
-		log_write("pmemfile_chdir(%p, \"%s\") = %ld",
-		    cwd_pool->pool, where.path, result);
-
-		pool_release(cwd_pool);
-
-		check_errno(result, SYS_chdir);
-	}
-
-	util_rwlock_unlock(&pmem_cwd_lock);
-
-	return result;
-}
-
-static long
-hook_fchdir(long fd)
-{
-	if (fd == AT_FDCWD)
-		return 0;
-
-	long result;
-
-	log_write("%s(%ld)", __func__, fd);
-
-	struct fd_association file = fd_ref(fd);
-
-	util_rwlock_wrlock(&pmem_cwd_lock);
-
-	if (!is_fda_null(&file)) {
-		struct pool_description *pool = file.pool;
-
-		pool_acquire(pool);
-
-		if (pmemfile_fchdir(pool->pool, file.file) == 0) {
-			cwd_pool = file.pool;
-			result = 0;
 		} else {
-			result = -errno;
+			result = pmemfile_vfd_chdir_pf(where.at_pool, file);
+			if (result != 0)
+				pmemfile_close(where.at_pool->pool, file);
 		}
 
-		log_write("pmemfile_fchdir(%p, %p) = %ld", pool->pool,
-				file.file, result);
+		pool_release(where.at_pool);
 
-		pool_release(pool);
-
-		check_errno(result, SYS_fchdir);
-	} else {
-		result = syscall_no_intercept(SYS_fchdir, fd);
-		if (result == 0)
-			cwd_pool = NULL;
+		result = check_errno(result, SYS_chdir);
 	}
 
-	util_rwlock_unlock(&pmem_cwd_lock);
+	pmemfile_vfd_unref(at);
 
-	fd_unref(fd, &file);
+	util_mutex_unlock(&cwd_mutex);
 
 	return result;
 }
 
 static long
-hook_getcwd(char *buf, size_t size)
+hook_pool_getcwd(struct pool_description *pool, char *buf, size_t size)
 {
-	if (cwd_pool == NULL)
+	if (pool == NULL)
 		return syscall_no_intercept(SYS_getcwd, buf, size);
 
-	size_t mlen = strlen(cwd_pool->mount_point);
+	size_t mlen = strlen(pool->mount_point);
 	if (mlen >= size)
 		return -ERANGE;
 
-	strcpy(buf, cwd_pool->mount_point);
+	strcpy(buf, pool->mount_point);
 
-	long ret;
+	long ret = 0;
 
-	pool_acquire(cwd_pool);
+	pool_acquire(pool);
 
-	if (pmemfile_getcwd(cwd_pool->pool, buf + mlen, size - mlen) == NULL)
+	if (pmemfile_getcwd(pool->pool, buf + mlen, size - mlen) == NULL)
 		ret = check_errno(-errno, SYS_getcwd);
-	else
-		ret = 0;
 
-	pool_release(cwd_pool);
+	pool_release(pool);
 
 	return ret;
 }
 
 static long
-hook_newfstatat(long fd, long arg0, long arg1, long arg2)
+hook_getcwd(char *buf, size_t size)
+{
+	util_mutex_lock(&cwd_mutex);
+	struct vfd_reference at = pmemfile_vfd_at_ref(AT_FDCWD);
+
+	long result = hook_pool_getcwd(at.pool, buf, size);
+
+	pmemfile_vfd_unref(at);
+	util_mutex_unlock(&cwd_mutex);
+
+	return result;
+}
+
+static long
+hook_newfstatat(int fd, long arg0, long arg1, long arg2)
 {
 	long ret;
 	struct resolved_path where;
 
-	struct fd_desc at = fd_fetch(fd);
+	struct vfd_reference at = pmemfile_vfd_at_ref(fd);
 
 	resolve_path(at, (const char *)arg0, &where,
 	    (arg2 & AT_SYMLINK_NOFOLLOW)
@@ -605,58 +407,53 @@ hook_newfstatat(long fd, long arg0, long arg1, long arg2)
 
 	if (where.error_code != 0) {
 		ret = where.error_code;
-	} else if (is_fda_null(&where.at.pmem_fda)) {
+	} else if (where.at_pool == NULL) {
 		ret = syscall_no_intercept(SYS_newfstatat,
-		    where.at.kernel_fd, where.path, arg1, arg2);
+		    where.at_kernel, where.path, arg1, arg2);
 	} else {
-		struct pool_description *pool = where.at.pmem_fda.pool;
+		pool_acquire(where.at_pool);
 
-		pool_acquire(pool);
-
-		int r = wrapper_pmemfile_fstatat(pool->pool,
-			where.at.pmem_fda.file,
-			where.path,
+		int r = wrapper_pmemfile_fstatat(where.at_pool->pool,
+			where.at_dir, where.path,
 			(pmemfile_stat_t *)arg1, (int)arg2);
 
-		pool_release(pool);
+		pool_release(where.at_pool);
 
 		ret = check_errno(r, SYS_newfstatat);
 	}
 
-	fd_release(&at);
+	pmemfile_vfd_unref(at);
 
 	return ret;
 }
 
 static long
-hook_faccessat(long fd, long path_arg, long mode)
+hook_faccessat(int fd, long path_arg, long mode)
 {
 	long ret;
 	struct resolved_path where;
 
-	struct fd_desc at = fd_fetch(fd);
+	struct vfd_reference at = pmemfile_vfd_ref(fd);
 
 	resolve_path(at, (const char *)path_arg, &where, NO_RESOLVE_LAST_SLINK);
 
 	if (where.error_code != 0) {
 		ret = where.error_code;
-	} else if (is_fda_null(&where.at.pmem_fda)) {
+	} else if (where.at_pool == NULL) {
 		ret = syscall_no_intercept(SYS_faccessat,
-		    where.at.kernel_fd, where.path, mode);
+		    where.at_kernel, where.path, mode);
 	} else {
-		struct pool_description *pool = where.at.pmem_fda.pool;
+		pool_acquire(where.at_pool);
 
-		pool_acquire(pool);
+		int r = wrapper_pmemfile_faccessat(where.at_pool->pool,
+			where.at_dir, where.path, (int)mode, 0);
 
-		int r = wrapper_pmemfile_faccessat(pool->pool,
-			where.at.pmem_fda.file, where.path, (int)mode, 0);
-
-		pool_release(pool);
+		pool_release(where.at_pool);
 
 		ret = check_errno(r, SYS_faccessat);
 	}
 
-	fd_release(&at);
+	pmemfile_vfd_unref(at);
 
 	return ret;
 }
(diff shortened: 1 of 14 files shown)


marcinslusarz:
FTR, I won't review this PR until it's rebased on top of master with #283 merged.

src/libpmemfile/vfd_table.c
```diff
@@ -0,0 +1,663 @@
...
+
+	*entry = (struct vfile_description) {
+		.pool = pool, .file = file,
+		.kernel_cwd_fd = -1,
+		.is_special_cwd_desc = false,
+		.ref_count = 1};
+
+	util_mutex_lock(&vfd_table_mutex);
+
+	vfd_table[new_vfd] = entry;
```
GBuella: Well, we might check it, yes.
This is a new fd from the kernel, so that would mean either that pmemfile_vfd_close is buggy, or the kernel is buggy.

GBuella: But if we assume those are OK, we could avoid the using the vfd_table_mutex here.

marcinslusarz: Yes, you are right, we don't need to check that.

src/libpmemfile/preload.c
```diff
@@ -413,155 +251,135 @@ hook_unlinkat(long fd, long path_arg, long flags)
...
-		result = syscall_no_intercept(SYS_chdir, where.path);
-		if (result == 0)
-			cwd_pool = NULL;
-	} else {
-		if (cwd_pool != where.at.pmem_fda.pool) {
-			cwd_pool = where.at.pmem_fda.pool;
-			syscall_no_intercept(SYS_chdir, cwd_pool->mount_point);
-		}
+	} else if (where.at_pool == NULL) {
+		long fd = syscall_no_intercept(SYS_open, path,
```
marcinslusarz: openat, where.at_kernel, where.path

src/libpmemfile/vfd_table.c
```diff
@@ -0,0 +1,663 @@
...
+	long result;
+	struct vfile_description *old_cwd_entry = NULL;
+
+	util_mutex_lock(&vfd_table_mutex);
+
+	if (is_in_vfd_table_range(vfd) && vfd_table[vfd] != NULL) {
+		vf_ref_count_inc(vfd_table[vfd]);
+		old_cwd_entry = cwd_entry;
+		cwd_entry = vfd_table[vfd];
+		result = 0;
```
marcinslusarz: where is pmemfile_fchdir?

src/libpmemfile/vfd_table.c
```diff
@@ -0,0 +1,663 @@
...
+	} else {
+		long new_fd = syscall_no_intercept(SYS_dup, vfd);
```
marcinslusarz: Please add comment explaining why are you duplicating fd and when it will be closed.

src/libpmemfile/preload.c
```diff
@@ -570,101 +388,109 @@ static long
...
+	if (file == NULL)
 		return check_errno(-errno, SYS_openat);
-	}
-
-	util_mutex_lock(&fd_table_mutex);
 
-	__sync_add_and_fetch(&fd_table[fd].ref_count, 1);
-	fd_table[fd].pmemfile.pool = where->at.pmem_fda.pool;
-	fd_table[fd].pmemfile.file = file;
+	long fd = pmemfile_vfd_assign(where->at_pool, file, where->path);
```
marcinslusarz: I think we should acquire file descriptor before calling pmemfile_open. With O_CREAT rolling back "open" would have to do more than just "close".

src/libpmemfile/vfd_table.c
```diff
@@ -0,0 +1,663 @@
...
+pmemfile_vfd_close(int vfd)
+{
+	struct vfile_description *entry = NULL;
+
+	util_mutex_lock(&vfd_table_mutex);
+
+	entry = vfd_table[vfd];
+	vfd_table[vfd] = NULL;
+
+	long result = syscall_no_intercept(SYS_close, vfd);
```
marcinslusarz: assert(!entry->is_special_cwd_desc); ?

sarahjelinek:
Reviewed 3 of 13 files at r1, 1 of 4 files at r2.
Review status: 3 of 13 files reviewed at latest revision, 22 unresolved discussions.

---

*[src/libpmemfile/path_resolve.c, line 74 at r3](https://reviewable.io:443/reviews/pmem/pmemfile/291#-KseorlR3GX3wZQOQgmL:-KseorlR3GX3wZQOQgmM:b-rkafym) ([raw file](https://github.com/pmem/pmemfile/blob/4c1bc46ec2833d0b4f8af1c0ffab56814f233355/src/libpmemfile/path_resolve.c#L74)):*
> ```C
> 	} else {
> 		int r = pmemfile_fstatat(result->at_pool->pool,
> 			result->at_dir,
> ```

Why do you automatically put a new line for each argument.  We should got to 80 chars and then indent. Same all through the code.

---

*[src/libpmemfile/preload.c, line 361 at r3](https://reviewable.io:443/reviews/pmem/pmemfile/291#-KsesiLgq4VGJlKM2KEu:-KsesiLgq4VGJlKM2KEv:b4mxodz) ([raw file](https://github.com/pmem/pmemfile/blob/4c1bc46ec2833d0b4f8af1c0ffab56814f233355/src/libpmemfile/preload.c#L361)):*
> ```C
> 
> static long
> hook_faccessat(int fd, long path_arg, long mode)
> ```

Shouldn't faccessat have 4 arguments?Last one is flags?

---

*[src/libpmemfile/preload.c, line 935 at r3](https://reviewable.io:443/reviews/pmem/pmemfile/291#-KsexeTwMgZJTe7hC2t-:-KsexeTwMgZJTe7hC2t0:bzf2l3b) ([raw file](https://github.com/pmem/pmemfile/blob/4c1bc46ec2833d0b4f8af1c0ffab56814f233355/src/libpmemfile/preload.c#L935)):*
> ```C
> 	}
> 
> 	return check_errno(r, sc);
> ```

'r' is an int here.  Argument to check_errno should be a long. Surprised your not getting a compiler error.  I isn't clear why you deleted 'long ret' and went to 'int r'. 

---

*[src/libpmemfile/vfd_table.c, line 287 at r3](https://reviewable.io:443/reviews/pmem/pmemfile/291#-KsfA1-uut-yXBTL5sMg:-KsfA1-uut-yXBTL5sMh:b-okwu0b) ([raw file](https://github.com/pmem/pmemfile/blob/4c1bc46ec2833d0b4f8af1c0ffab56814f233355/src/libpmemfile/vfd_table.c#L287)):*
> ```C
>  * as different, and all three happen while holding the vfd_table_mutex.
>  *
>  * Important: dup2 must atomic from the user's point of view.
> ```

s/must/must be

---

*[src/libpmemfile/vfd_table.h, line 2 at r2](https://reviewable.io:443/reviews/pmem/pmemfile/291#-KsEAvbX9xlhqBXpENei:-KsEAvbX9xlhqBXpENej:bih59i3) ([raw file](https://github.com/pmem/pmemfile/blob/5b5bdd8a8b213029aadd681614d34fe108c0d0af/src/libpmemfile/vfd_table.h#L2)):*
> ```objc
> /*
>  * Copyright 2016-2017, Intel Corporation
> ```

Need only 2017.

---

*[src/libpmemfile/vfd_table.h, line 62 at r2](https://reviewable.io:443/reviews/pmem/pmemfile/291#-KsEBAvcTAAmXd1-h5B7:-KsEBAvcTAAmXd1-h5B8:butkcri) ([raw file](https://github.com/pmem/pmemfile/blob/5b5bdd8a8b213029aadd681614d34fe108c0d0af/src/libpmemfile/vfd_table.h#L62)):*
> ```objc
> 
> long pmemfile_vfd_chdir_pf(struct pool_description *pool,
> 				struct pmemfile_file *file);
> ```

Should be consistent with tabs. This one has 4, the one below has 3 for continuation.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/291)*

marcinslusarz:
*[src/libpmemfile/preload.c, line 361 at r3](https://reviewable.io:443/reviews/pmem/pmemfile/291#-KsesiLgq4VGJlKM2KEu:-KshHfPiEBge9oSGTU0D:bv8yqof) ([raw file](https://github.com/pmem/pmemfile/blob/4c1bc46ec2833d0b4f8af1c0ffab56814f233355/src/libpmemfile/preload.c#L361)):*

Shouldn't faccessat have 4 arguments?Last one is flags?
</blockquote></details>

From faccessat manpage:
"   C library/kernel differences
       The  raw  faccessat() system call takes only the first three arguments.  The AT_EACCESS and AT_SYMLINK_NOFOLLOW flags are actually implemented within the glibc wrapper function for faccessat().  If
       either of these flags is specified, then the wrapper function employs fstatat(2) to determine access permissions."

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/291)*

krzycz:
Reviewed 9 of 13 files at r1, 2 of 4 files at r2, 3 of 3 files at r3.
Review status: all files reviewed at latest revision, 22 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/291)*

GBuella:
Review status: 10 of 13 files reviewed at latest revision, 22 unresolved discussions.

---

*[src/libpmemfile/preload.c, line 1028 at r2](https://reviewable.io:443/reviews/pmem/pmemfile/291#-KqbfDlP2bvWwh_vWWnX-r2-1028:-Ksi2B80ah9-ZaiKFvS6:b-896fix) ([raw file](https://github.com/pmem/pmemfile/blob/5b5bdd8a8b213029aadd681614d34fe108c0d0af/src/libpmemfile/preload.c#L1028)):*

ret = syscall_no_intercept(....
</blockquote></details>

Done.

---

*[src/libpmemfile/preload.c, line 2076 at r2](https://reviewable.io:443/reviews/pmem/pmemfile/291#-KqbfDlP2bvWwh_vWWnX-r2-2076:-Ksi2CRcNdCeWsmp69l4:b-896fix) ([raw file](https://github.com/pmem/pmemfile/blob/5b5bdd8a8b213029aadd681614d34fe108c0d0af/src/libpmemfile/preload.c#L2076)):*

It should fit in previous line.
</blockquote></details>

Done.

---

*[src/libpmemfile/vfd_table.c, line 435 at r2](https://reviewable.io:443/reviews/pmem/pmemfile/291#-KqbfDlP2bvWwh_vWWnk-r2-435:-Ksi2Kc0P71nghNJmx1Y:b-896fix) ([raw file](https://github.com/pmem/pmemfile/blob/5b5bdd8a8b213029aadd681614d34fe108c0d0af/src/libpmemfile/vfd_table.c#L435)):*

Please don't use this syntax. It confuses Eclipse in a non-obvious way :(
</blockquote></details>

Done.

---

*[src/libpmemfile/vfd_table.c, line 287 at r3](https://reviewable.io:443/reviews/pmem/pmemfile/291#-KsfA1-uut-yXBTL5sMg:-Ksi2HybI9eaqKMTM1HR:b-896fix) ([raw file](https://github.com/pmem/pmemfile/blob/4c1bc46ec2833d0b4f8af1c0ffab56814f233355/src/libpmemfile/vfd_table.c#L287)):*

s/must/must be
</blockquote></details>

Done.

---

*[src/libpmemfile/vfd_table.h, line 2 at r2](https://reviewable.io:443/reviews/pmem/pmemfile/291#-KsEAvbX9xlhqBXpENei:-Ksi2NfQxh5fEtmOLaFd:b-896fix) ([raw file](https://github.com/pmem/pmemfile/blob/5b5bdd8a8b213029aadd681614d34fe108c0d0af/src/libpmemfile/vfd_table.h#L2)):*

Need only 2017.
</blockquote></details>

Done.

---

*[src/libpmemfile/vfd_table.h, line 62 at r2](https://reviewable.io:443/reviews/pmem/pmemfile/291#-KsEBAvcTAAmXd1-h5B7:-Ksi2OPBQWoPI7c5fK1Y:b-896fix) ([raw file](https://github.com/pmem/pmemfile/blob/5b5bdd8a8b213029aadd681614d34fe108c0d0af/src/libpmemfile/vfd_table.h#L62)):*

Should be consistent with tabs. This one has 4, the one below has 3 for continuation.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/291)*

src/libpmemfile/preload.c
GBuella: That might make things a bit more complicated.
1) acquire an fd
2) call pmemfile_openat
3) associate the fd with the pmemfile_file

I would have to split pmemfile_vfd_assign to two steps.
Should that be done? Is it worth it?

GBuella: Also, probably we would have to hold the vfd_table_mutex during the whole thing.

GBuella:
Review status: 9 of 13 files reviewed at latest revision, 22 unresolved discussions.

---

*[src/libpmemfile/path_resolve.c, line 74 at r3](https://reviewable.io:443/reviews/pmem/pmemfile/291#-KseorlR3GX3wZQOQgmL:-Ksi5ZH47IsP3pfPCatA:b-896fix) ([raw file](https://github.com/pmem/pmemfile/blob/4c1bc46ec2833d0b4f8af1c0ffab56814f233355/src/libpmemfile/path_resolve.c#L74)):*

Why do you automatically put a new line for each argument.  We should got to 80 chars and then indent. Same all through the code.
</blockquote></details>

Done.

---

*[src/libpmemfile/preload.c, line 267 at r2](https://reviewable.io:443/reviews/pmem/pmemfile/291#-KqbfDlP2bvWwh_vWWnX-r2-267:-Ksi5_1vyFR2rsKqarBV:b-896fix) ([raw file](https://github.com/pmem/pmemfile/blob/5b5bdd8a8b213029aadd681614d34fe108c0d0af/src/libpmemfile/preload.c#L267)):*

openat, where.at_kernel, where.path
</blockquote></details>

Done.

---

*[src/libpmemfile/preload.c, line 2086 at r2](https://reviewable.io:443/reviews/pmem/pmemfile/291#-KqbfDlP2bvWwh_vWWnX-r2-2086:-Ksi5fOhiaizzvRaUHt7:b-896fix) ([raw file](https://github.com/pmem/pmemfile/blob/5b5bdd8a8b213029aadd681614d34fe108c0d0af/src/libpmemfile/preload.c#L2086)):*

This comment should probably be in function description.
</blockquote></details>

Done.

---

*[src/libpmemfile/preload.c, line 935 at r3](https://reviewable.io:443/reviews/pmem/pmemfile/291#-KsexeTwMgZJTe7hC2t-:-Ksi5oTYeXR0_2m8ZPP5:b-dlgc5q) ([raw file](https://github.com/pmem/pmemfile/blob/4c1bc46ec2833d0b4f8af1c0ffab56814f233355/src/libpmemfile/preload.c#L935)):*

'r' is an int here.  Argument to check_errno should be a long. Surprised your not getting a compiler error.  I isn't clear why you deleted 'long ret' and went to 'int r'. 
</blockquote></details>

`pmemfile_utimensat` returns `int`, and libc `utimensat` returns `int`, the kernel also returns `int` I  believe.
As of now, all these functions return a `long`, which I just put into rax.
Probably I will eventually match libc return types in these wrapper functions.

Also, `long` must be able to represent anything that an `int` can represent according to the standard, so this cast should never result in an error.
Here, long refers to a 64 bit integer, int to a 32 bit integer.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/291)*

marcinslusarz:
*[src/libpmemfile/preload.c, line 484 at r2](https://reviewable.io:443/reviews/pmem/pmemfile/291#-KqbfDlP2bvWwh_vWWnX-r2-484:-Ksi9Tfq37_zVo2_aPAZ:b-va2qgo) ([raw file](https://github.com/pmem/pmemfile/blob/5b5bdd8a8b213029aadd681614d34fe108c0d0af/src/libpmemfile/preload.c#L484)):*

Also, probably we would have to hold the vfd_table_mutex during the whole thing.
</blockquote></details>

Can't you acquire file descriptor without registering it anywhere and associate pmemfile_file with already generated fd?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/291)*

marcinslusarz:
Reviewed 8 of 13 files at r1, 1 of 4 files at r2, 1 of 4 files at r4.
Review status: 10 of 13 files reviewed at latest revision, 16 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/291)*

krzycz:
Reviewed 4 of 4 files at r4.
Review status: all files reviewed at latest revision, 16 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/291)*

GBuella:
Review status: all files reviewed at latest revision, 11 unresolved discussions.

---

*[src/libpmemfile/preload.c, line 484 at r2](https://reviewable.io:443/reviews/pmem/pmemfile/291#-KqbfDlP2bvWwh_vWWnX-r2-484:-KsmUVn_jXPhCon0gFGf:bqjfn78) ([raw file](https://github.com/pmem/pmemfile/blob/5b5bdd8a8b213029aadd681614d34fe108c0d0af/src/libpmemfile/preload.c#L484)):*

Can't you acquire file descriptor without registering it anywhere and associate pmemfile_file with already generated fd?
</blockquote></details>

You are right, I was silly, no one else acquire the same fd once it is acquired.
I'll get to it

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/291)*

GBuella:
Review status: all files reviewed at latest revision, 11 unresolved discussions.

---

*[src/libpmemfile/vfd_table.c, line 83 at r2](https://reviewable.io:443/reviews/pmem/pmemfile/291#-KqbfDlP2bvWwh_vWWnk-r2-83:-KsmV7oyB8S6fjyKRGJq:b-ma4vcr) ([raw file](https://github.com/pmem/pmemfile/blob/5b5bdd8a8b213029aadd681614d34fe108c0d0af/src/libpmemfile/vfd_table.c#L83)):*

Why is it twice the size of vfd_table?
</blockquote></details>

I recall my reasoning was something like:
The maximum number of files open is the number of entries in the vfd_table plus the number of threads holding a file description that is already closed (using the deferred close).
But of course the number of threads can be different.

I think I can make this table smaller, and dynamically reallocate it if more is ever needed.
What do you think?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/291)*

marcinslusarz:
Review status: all files reviewed at latest revision, 10 unresolved discussions.

---

*[src/libpmemfile/vfd_table.c, line 83 at r2](https://reviewable.io:443/reviews/pmem/pmemfile/291#-KqbfDlP2bvWwh_vWWnk-r2-83:-KsmXgdR1Pj1-j2gLe7C:b-xqv6s7) ([raw file](https://github.com/pmem/pmemfile/blob/5b5bdd8a8b213029aadd681614d34fe108c0d0af/src/libpmemfile/vfd_table.c#L83)):*

I recall my reasoning was something like:
The maximum number of files open is the number of entries in the vfd_table plus the number of threads holding a file description that is already closed (using the deferred close).
But of course the number of threads can be different.

I think I can make this table smaller, and dynamically reallocate it if more is ever needed.
What do you think?
</blockquote></details>

Can you have more free slots than the number of elements in the array with real entries?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/291)*

GBuella:
Review status: all files reviewed at latest revision, 10 unresolved discussions.

---

*[src/libpmemfile/vfd_table.c, line 83 at r2](https://reviewable.io:443/reviews/pmem/pmemfile/291#-KqbfDlP2bvWwh_vWWnk-r2-83:-KsmoDXy8kAsQ0n3AjRt:bru04br) ([raw file](https://github.com/pmem/pmemfile/blob/5b5bdd8a8b213029aadd681614d34fe108c0d0af/src/libpmemfile/vfd_table.c#L83)):*

Can you have more free slots than the number of elements in the array with real entries?
</blockquote></details>

Yes, the number of free slots is not limited by that.
A free slot is used up when opening a file, and a slot is freed up when its refcount reaches zero.
That array with real entries only contains pointers to slots, and constitutes one reference. Other references are by threads using the file.
Someone can ask for a free slot whenever there is appropriate place in the vfd_table - and it would be nice to have a free slot available.

I guess we could just return *EMFILE*, or *ENOMEM*, and be done with it. It would be a pretty rare scenario.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/291)*

GBuella:
Review status: 12 of 13 files reviewed at latest revision, 10 unresolved discussions.

---

*[src/libpmemfile/vfd_table.c, line 247 at r2](https://reviewable.io:443/reviews/pmem/pmemfile/291#-KqbfDlP2bvWwh_vWWnk-r2-247:-Ksmqi1bk-gWkzRBA2S7:b-896fix) ([raw file](https://github.com/pmem/pmemfile/blob/5b5bdd8a8b213029aadd681614d34fe108c0d0af/src/libpmemfile/vfd_table.c#L247)):*

releases
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/291)*

marcinslusarz:
*[src/libpmemfile/vfd_table.c, line 83 at r2](https://reviewable.io:443/reviews/pmem/pmemfile/291#-KqbfDlP2bvWwh_vWWnk-r2-83:-KsmyfVzsOaK3rrdQFhi:bpoyk7m) ([raw file](https://github.com/pmem/pmemfile/blob/5b5bdd8a8b213029aadd681614d34fe108c0d0af/src/libpmemfile/vfd_table.c#L83)):*

Yes, the number of free slots is not limited by that.
A free slot is used up when opening a file, and a slot is freed up when its refcount reaches zero.
That array with real entries only contains pointers to slots, and constitutes one reference. Other references are by threads using the file.
Someone can ask for a free slot whenever there is appropriate place in the vfd_table - and it would be nice to have a free slot available.

I guess we could just return *EMFILE*, or *ENOMEM*, and be done with it. It would be a pretty rare scenario.
</blockquote></details>

So you are saying we can have vfile_description that is not in vfd_table? How is that possible? Each vfile_description should have at least one associated file descriptor. What am I missing?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/291)*

krzycz:
Reviewed 1 of 1 files at r5.
Review status: all files reviewed at latest revision, 10 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/291)*

GBuella:
Review status: all files reviewed at latest revision, 10 unresolved discussions.

---

*[src/libpmemfile/vfd_table.c, line 83 at r2](https://reviewable.io:443/reviews/pmem/pmemfile/291#-KqbfDlP2bvWwh_vWWnk-r2-83:-Ksn3j2FmFrSSBk5mjLZ:b-vu2h70) ([raw file](https://github.com/pmem/pmemfile/blob/5b5bdd8a8b213029aadd681614d34fe108c0d0af/src/libpmemfile/vfd_table.c#L83)):*

So you are saying we can have vfile_description that is not in vfd_table? How is that possible? Each vfile_description should have at least one associated file descriptor. What am I missing?
</blockquote></details>

No, they don't need to have an associated fd number. That can happen e.g. when one thread is reading from a file, and another thread closes the fd. The close syscall is serviced by removing the pointer from the vfd_table, and decreasing the refcount. The reading thread meanwhile keeps on reading, and decreases the refcount once it is done reading - not caring about the vfd_table.
A slot is only returned to the free slot list, and the underlying pmemfile_file is only closed when the refcount reaches zero.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/291)*

marcinslusarz:
Review status: all files reviewed at latest revision, 9 unresolved discussions, some commit checks failed.

---

*[src/libpmemfile/vfd_table.c, line 83 at r2](https://reviewable.io:443/reviews/pmem/pmemfile/291#-KqbfDlP2bvWwh_vWWnk-r2-83:-KsnDRuPpL6Ec7znTNws:b6lv1kv) ([raw file](https://github.com/pmem/pmemfile/blob/5b5bdd8a8b213029aadd681614d34fe108c0d0af/src/libpmemfile/vfd_table.c#L83)):*

No, they don't need to have an associated fd number. That can happen e.g. when one thread is reading from a file, and another thread closes the fd. The close syscall is serviced by removing the pointer from the vfd_table, and decreasing the refcount. The reading thread meanwhile keeps on reading, and decreases the refcount once it is done reading - not caring about the vfd_table.
A slot is only returned to the free slot list, and the underlying pmemfile_file is only closed when the refcount reaches zero.
</blockquote></details>

Ok, I understand it now, thanks. I think failing to open/create file when there's not enough slots is the saner option, because what you are trying to handle is corner case (there are pending operations for closed file descriptors) of a corner case (almost all fds are allocated). It's not worth it.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/291)*

GBuella:
Review status: 5 of 13 files reviewed at latest revision, 9 unresolved discussions.

---

*[src/libpmemfile/vfd_table.c, line 83 at r2](https://reviewable.io:443/reviews/pmem/pmemfile/291#-KqbfDlP2bvWwh_vWWnk-r2-83:-KtBgXxz7drfnvnj2uFN:b3osmqt) ([raw file](https://github.com/pmem/pmemfile/blob/5b5bdd8a8b213029aadd681614d34fe108c0d0af/src/libpmemfile/vfd_table.c#L83)):*

Ok, I understand it now, thanks. I think failing to open/create file when there's not enough slots is the saner option, because what you are trying to handle is corner case (there are pending operations for closed file descriptors) of a corner case (almost all fds are allocated). It's not worth it.
</blockquote></details>

I changed the logic around it a bit.
Now fetch_free_file_slot returns NULL, if there no more free slots available.

---

*[src/libpmemfile/vfd_table.c, line 105 at r2](https://reviewable.io:443/reviews/pmem/pmemfile/291#-KqbfDlP2bvWwh_vWWnk-r2-105:-KtBgiwWr07xmyIHT1j6:b-xszchr) ([raw file](https://github.com/pmem/pmemfile/blob/5b5bdd8a8b213029aadd681614d34fe108c0d0af/src/libpmemfile/vfd_table.c#L105)):*
> free_slot_count
There is a variable called `free_slot_count` now.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/291)*

marcinslusarz:
Review status: 5 of 13 files reviewed at latest revision, 8 unresolved discussions.

---

*[src/libpmemfile/vfd_table.c, line 372 at r2](https://reviewable.io:443/reviews/pmem/pmemfile/291#-KqbfDlP2bvWwh_vWWnk-r2-372:-KtBtDo-5B9GSnVrcGtw:b-p0jbya) ([raw file](https://github.com/pmem/pmemfile/blob/5b5bdd8a8b213029aadd681614d34fe108c0d0af/src/libpmemfile/vfd_table.c#L372)):*

Done.
</blockquote></details>

You don't need to use parentheses...

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/291)*

marcinslusarz:
Reviewed 1 of 4 files at r4, 6 of 8 files at r6.
Review status: 11 of 13 files reviewed at latest revision, 14 unresolved discussions.

---

*[src/libpmemfile/preload.c, line 604 at r2](https://reviewable.io:443/reviews/pmem/pmemfile/291#-KqbfDlP2bvWwh_vWWnX-r2-604:-KtC0P-gFj4pElNwFzV6:b-1j44rb) ([raw file](https://github.com/pmem/pmemfile/blob/5b5bdd8a8b213029aadd681614d34fe108c0d0af/src/libpmemfile/preload.c#L604)):*

I just realized that what we have on master is buggy - truncate has only 2 arguments... We probably should set up tests for files outside of pmemfile pool.
</blockquote></details>

Version from master always fails, but the new one may succeed silently corrupting something. I'd prefer the former.

---

*[src/libpmemfile/preload.c, line 338 at r6](https://reviewable.io:443/reviews/pmem/pmemfile/291#-KtBzyvf7UHoH9rHBZ4Y:-KtBzyvgkCKXlJrStiPm:bv8qe8j) ([raw file](https://github.com/pmem/pmemfile/blob/e43dfb3b2ebe2571262e77a43faa4e3cd45e7d20/src/libpmemfile/preload.c#L338)):*
> ```C
> 					O_DIRECTORY | O_PATH | O_NOCTTY);
> 
> 		pool_release(where.at_pool);
> ```

I think you should move it to after pmemfile_close.

---

*[src/libpmemfile/preload.c, line 939 at r6](https://reviewable.io:443/reviews/pmem/pmemfile/291#-KtC1NJk2HRT3AgTym7d:-KtC1NJk2HRT3AgTym7e:btj05i0) ([raw file](https://github.com/pmem/pmemfile/blob/e43dfb3b2ebe2571262e77a43faa4e3cd45e7d20/src/libpmemfile/preload.c#L939)):*
> ```C
> 				"pmemfile_futimesat(%p, %p, \"%s\", [%ld,%ld,%ld,%ld]) = %d",
> 			    (void *) where.at_pool->pool,
> 			    (void *) where.at_dir, where.path,
> ```

Please remove those casts.

---

*[src/libpmemfile/preload.c, line 946 at r6](https://reviewable.io:443/reviews/pmem/pmemfile/291#-KtC1kiTVpxX0X9jNDzY:-KtC1kiTVpxX0X9jNDzZ:bs4q9op) ([raw file](https://github.com/pmem/pmemfile/blob/e43dfb3b2ebe2571262e77a43faa4e3cd45e7d20/src/libpmemfile/preload.c#L946)):*
> ```C
> 				"pmemfile_futimesat(%p, %p, \"%s\", NULL) = %d",
> 			    (void *) where.at_pool->pool,
> 			    (void *) where.at_dir, where.path, r);
> ```

Ditto.

---

*[src/libpmemfile/preload.c, line 1017 at r6](https://reviewable.io:443/reviews/pmem/pmemfile/291#-KtC1r1tQPlWKt1hPIB3:-KtC1r1tQPlWKt1hPIB4:b3kqx) ([raw file](https://github.com/pmem/pmemfile/blob/e43dfb3b2ebe2571262e77a43faa4e3cd45e7d20/src/libpmemfile/preload.c#L1017)):*
> ```C
> 				"pmemfile_futimens(%p, %p, [%ld,%ld,%ld,%ld]) = %d",
> 				(void *) where.at_pool->pool,
> 				(void *) where.at_dir,
> ```

.

---

*[src/libpmemfile/preload.c, line 1024 at r6](https://reviewable.io:443/reviews/pmem/pmemfile/291#-KtC1ruez3WdEGCEYqCg:-KtC1rufk6gq046aPzWp:b3kqx) ([raw file](https://github.com/pmem/pmemfile/blob/e43dfb3b2ebe2571262e77a43faa4e3cd45e7d20/src/libpmemfile/preload.c#L1024)):*
> ```C
> 				"pmemfile_futimens(%p, %p, NULL) = %d",
> 				(void *) where.at_pool->pool,
> 				(void *) where.at_dir, r);
> ```

.

---

*[src/libpmemfile/preload.c, line 1051 at r6](https://reviewable.io:443/reviews/pmem/pmemfile/291#-KtC2AqQuTZnXEsgGxWk:-KtC2AqRLcp1wSmVFXHa:btqe8v5) ([raw file](https://github.com/pmem/pmemfile/blob/e43dfb3b2ebe2571262e77a43faa4e3cd45e7d20/src/libpmemfile/preload.c#L1051)):*
> ```C
> 			    (void *) where.at_pool->pool,
> 			    (void *) where.at_dir,
> 			    where.path, flags, r);
> ```

You basically forced your version, including reversion of format string...

---

*[src/libpmemfile/preload.c, line 2328 at r6](https://reviewable.io:443/reviews/pmem/pmemfile/291#-KtC7Y9vZXpw0zayvRJi:-KtC7Y9vZXpw0zayvRJj:bf7jdzh) ([raw file](https://github.com/pmem/pmemfile/blob/e43dfb3b2ebe2571262e77a43faa4e3cd45e7d20/src/libpmemfile/preload.c#L2328)):*
> ```C
> 	if (file == NULL)
> 		exit_with_msg(PMEMFILE_PRELOAD_EXIT_POOL_OPEN_FAILED,
> 				"!chdir into pmemfile_pool");
> ```

"!opening cwd on pmemfile_pool"

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/291)*

GBuella:
Review status: 11 of 13 files reviewed at latest revision, 14 unresolved discussions.

---

*[src/libpmemfile/preload.c, line 604 at r2](https://reviewable.io:443/reviews/pmem/pmemfile/291#-KqbfDlP2bvWwh_vWWnX-r2-604:-KtCCmWGrzjNNzOENVB7:ben5axo) ([raw file](https://github.com/pmem/pmemfile/blob/5b5bdd8a8b213029aadd681614d34fe108c0d0af/src/libpmemfile/preload.c#L604)):*

Version from master always fails, but the new one may succeed silently corrupting something. I'd prefer the former.
</blockquote></details>

I added some checks, this should make the situation a little better.

---

*[src/libpmemfile/preload.c, line 338 at r6](https://reviewable.io:443/reviews/pmem/pmemfile/291#-KtBzyvf7UHoH9rHBZ4Y:-KtCCjSDlhdLH8h1fXuv:b-896fix) ([raw file](https://github.com/pmem/pmemfile/blob/e43dfb3b2ebe2571262e77a43faa4e3cd45e7d20/src/libpmemfile/preload.c#L338)):*

I think you should move it to after pmemfile_close.
</blockquote></details>

Done.

---

*[src/libpmemfile/preload.c, line 939 at r6](https://reviewable.io:443/reviews/pmem/pmemfile/291#-KtC1NJk2HRT3AgTym7d:-KtCCtsi4OTuytfHw3V0:b-896fix) ([raw file](https://github.com/pmem/pmemfile/blob/e43dfb3b2ebe2571262e77a43faa4e3cd45e7d20/src/libpmemfile/preload.c#L939)):*

Please remove those casts.
</blockquote></details>

Done.

---

*[src/libpmemfile/preload.c, line 946 at r6](https://reviewable.io:443/reviews/pmem/pmemfile/291#-KtC1kiTVpxX0X9jNDzY:-KtCCuB0ve55rtAspMpe:b-896fix) ([raw file](https://github.com/pmem/pmemfile/blob/e43dfb3b2ebe2571262e77a43faa4e3cd45e7d20/src/libpmemfile/preload.c#L946)):*

Ditto.
</blockquote></details>

Done.

---

*[src/libpmemfile/preload.c, line 1017 at r6](https://reviewable.io:443/reviews/pmem/pmemfile/291#-KtC1r1tQPlWKt1hPIB3:-KtCCudHoqu7yoOF596v:b-896fix) ([raw file](https://github.com/pmem/pmemfile/blob/e43dfb3b2ebe2571262e77a43faa4e3cd45e7d20/src/libpmemfile/preload.c#L1017)):*

.
</blockquote></details>

Done.

---

*[src/libpmemfile/preload.c, line 1024 at r6](https://reviewable.io:443/reviews/pmem/pmemfile/291#-KtC1ruez3WdEGCEYqCg:-KtCCuxfdES3cqP1XApM:b-896fix) ([raw file](https://github.com/pmem/pmemfile/blob/e43dfb3b2ebe2571262e77a43faa4e3cd45e7d20/src/libpmemfile/preload.c#L1024)):*

.
</blockquote></details>

Done.

---

*[src/libpmemfile/preload.c, line 1051 at r6](https://reviewable.io:443/reviews/pmem/pmemfile/291#-KtC2AqQuTZnXEsgGxWk:-KtCCv_TLbqYtFN2ct_h:b-896fix) ([raw file](https://github.com/pmem/pmemfile/blob/e43dfb3b2ebe2571262e77a43faa4e3cd45e7d20/src/libpmemfile/preload.c#L1051)):*

You basically forced your version, including reversion of format string...
</blockquote></details>

Done.

---

*[src/libpmemfile/preload.c, line 2328 at r6](https://reviewable.io:443/reviews/pmem/pmemfile/291#-KtC7Y9vZXpw0zayvRJi:-KtCCwOz_O91D7_7jAx3:b-896fix) ([raw file](https://github.com/pmem/pmemfile/blob/e43dfb3b2ebe2571262e77a43faa4e3cd45e7d20/src/libpmemfile/preload.c#L2328)):*

"!opening cwd on pmemfile_pool"
</blockquote></details>

Done.

---

*[src/libpmemfile/vfd_table.c, line 625 at r2](https://reviewable.io:443/reviews/pmem/pmemfile/291#-KqbfDlP2bvWwh_vWWnk-r2-625:-KtCCzfWduo3KiHpfun3:b-wskc72) ([raw file](https://github.com/pmem/pmemfile/blob/5b5bdd8a8b213029aadd681614d34fe108c0d0af/src/libpmemfile/vfd_table.c#L625)):*

where is pmemfile_fchdir?
</blockquote></details>

I forgot about it!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/291)*

marcinslusarz:
Review status: 11 of 13 files reviewed at latest revision, 10 unresolved discussions.

---

*[src/libpmemfile/preload.c, line 1051 at r6](https://reviewable.io:443/reviews/pmem/pmemfile/291#-KtC2AqQuTZnXEsgGxWk:-KtCFp2CrSY9FvGEIcqh:buphna0) ([raw file](https://github.com/pmem/pmemfile/blob/e43dfb3b2ebe2571262e77a43faa4e3cd45e7d20/src/libpmemfile/preload.c#L1051)):*

Done.
</blockquote></details>

Not really. cstyle allows lines with strings to bypass 80 characters limit.

---

*[src/libpmemfile/vfd_table.c, line 180 at r6](https://reviewable.io:443/reviews/pmem/pmemfile/291#-KtCCtaTN_MxrWzT4Hdy:-KtCCtaUnqHQajsKjx-k:b3own6q) ([raw file](https://github.com/pmem/pmemfile/blob/e43dfb3b2ebe2571262e77a43faa4e3cd45e7d20/src/libpmemfile/vfd_table.c#L180)):*
> ```C
> /*
>  * pmemfile_vfd_ref -- return a vfd_reference, which is either actually
>  * a refernce to vfile_description entry, or just holding an fd handled by
> ```

s/refernce/reference/

---

*[src/libpmemfile/vfd_table.c, line 390 at r6](https://reviewable.io:443/reviews/pmem/pmemfile/291#-KtCDdVqKPHLz03TeN4m:-KtCDdVrFmJc-8ummZj9:b3kqs) ([raw file](https://github.com/pmem/pmemfile/blob/e43dfb3b2ebe2571262e77a43faa4e3cd45e7d20/src/libpmemfile/vfd_table.c#L390)):*
> ```C
> /*
>  * pmemfile_vfd_close -- remove a reference from the vfd_table array (if
>  * there was one at vfd_table[vfd].
> ```

)

---

*[src/libpmemfile/vfd_table.c, line 498 at r6](https://reviewable.io:443/reviews/pmem/pmemfile/291#-KtCDyJNp7rUzkwlSF5h:-KtCDyJOuP5E_zQOPEpf:b2nhipw) ([raw file](https://github.com/pmem/pmemfile/blob/e43dfb3b2ebe2571262e77a43faa4e3cd45e7d20/src/libpmemfile/vfd_table.c#L498)):*
> ```C
> 		} else {
> 			result = -ENOMEM;
> 			/* Good luck making that line green in coverity! */
> ```

LOL

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/291)*

GBuella:
Review status: 11 of 13 files reviewed at latest revision, 9 unresolved discussions.

---

*[src/libpmemfile/preload.c, line 1051 at r6](https://reviewable.io:443/reviews/pmem/pmemfile/291#-KtC2AqQuTZnXEsgGxWk:-KtCJ-aR68oa0iftZtk1:b-896fix) ([raw file](https://github.com/pmem/pmemfile/blob/e43dfb3b2ebe2571262e77a43faa4e3cd45e7d20/src/libpmemfile/preload.c#L1051)):*

Not really. cstyle allows lines with strings to bypass 80 characters limit.
</blockquote></details>

Done.

---

*[src/libpmemfile/vfd_table.c, line 450 at r2](https://reviewable.io:443/reviews/pmem/pmemfile/291#-KqbfDlP2bvWwh_vWWnk-r2-450:-KtCJA2DM2PRCnJSN_y3:b-uorc4w) ([raw file](https://github.com/pmem/pmemfile/blob/5b5bdd8a8b213029aadd681614d34fe108c0d0af/src/libpmemfile/vfd_table.c#L450)):*

This race could happen only if we would replace memory behind cwd_entry pointer. Since we are replacing pointer value and all accesses to this variable are under vfd_table_mutex there should be no race. Am I missing something?
</blockquote></details>

I removed this function.
The `cwd_entry` variable is always accessed under a lock, as you pointed out.

---

*[src/libpmemfile/vfd_table.c, line 180 at r6](https://reviewable.io:443/reviews/pmem/pmemfile/291#-KtCCtaTN_MxrWzT4Hdy:-KtCJ26U-lzGyRWxlc0w:b-896fix) ([raw file](https://github.com/pmem/pmemfile/blob/e43dfb3b2ebe2571262e77a43faa4e3cd45e7d20/src/libpmemfile/vfd_table.c#L180)):*

s/refernce/reference/
</blockquote></details>

Done.

---

*[src/libpmemfile/vfd_table.c, line 390 at r6](https://reviewable.io:443/reviews/pmem/pmemfile/291#-KtCDdVqKPHLz03TeN4m:-KtCJ3aAM9OAVoUkWh-N:b-896fix) ([raw file](https://github.com/pmem/pmemfile/blob/e43dfb3b2ebe2571262e77a43faa4e3cd45e7d20/src/libpmemfile/vfd_table.c#L390)):*

)
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/291)*

GBuella:
Review status: 11 of 13 files reviewed at latest revision, 9 unresolved discussions.

---

*[src/libpmemfile/vfd_table.c, line 625 at r2](https://reviewable.io:443/reviews/pmem/pmemfile/291#-KqbfDlP2bvWwh_vWWnk-r2-625:-KtCMo6rNNIrlGf5_H5e:b-896fix) ([raw file](https://github.com/pmem/pmemfile/blob/5b5bdd8a8b213029aadd681614d34fe108c0d0af/src/libpmemfile/vfd_table.c#L625)):*

I forgot about it!
</blockquote></details>

Done.

---

*[src/libpmemfile/vfd_table.c, line 627 at r2](https://reviewable.io:443/reviews/pmem/pmemfile/291#-KqbfDlP2bvWwh_vWWnk-r2-627:-KtCMgnpehp73Gnt9FOW:b-896fix) ([raw file](https://github.com/pmem/pmemfile/blob/5b5bdd8a8b213029aadd681614d34fe108c0d0af/src/libpmemfile/vfd_table.c#L627)):*

Please add comment explaining why are you duplicating fd and when it will be closed.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/291)*

GBuella:
Review status: 10 of 13 files reviewed at latest revision, 9 unresolved discussions.

---

*[src/libpmemfile/preload.c, line 484 at r2](https://reviewable.io:443/reviews/pmem/pmemfile/291#-KqbfDlP2bvWwh_vWWnX-r2-484:-KtCWIKHm5c0l-mOMpz1:b-896fix) ([raw file](https://github.com/pmem/pmemfile/blob/5b5bdd8a8b213029aadd681614d34fe108c0d0af/src/libpmemfile/preload.c#L484)):*

You are right, I was silly, no one else acquire the same fd once it is acquired.
I'll get to it
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/291)*

marcinslusarz:
Review status: 10 of 13 files reviewed at latest revision, 4 unresolved discussions.

---

*[src/libpmemfile/vfd_table.c, line 623 at r7](https://reviewable.io:443/reviews/pmem/pmemfile/291#-KtGfMNqyszfmylFyUbB:-KtGfMNqyszfmylFyUbC:b-r3n9vi) ([raw file](https://github.com/pmem/pmemfile/blob/14b4d9ac4ddd3d49e4a739599d67521d6ec6a2dc/src/libpmemfile/vfd_table.c#L623)):*
> ```C
> 	} else {
> 		/*
> 		 * Acquire a new fd keep in the cwd_entry. This will be
> ```

"fd to keep"?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/291)*

marcinslusarz:
Review status: 10 of 13 files reviewed at latest revision, 3 unresolved discussions.

---

*[src/libpmemfile/preload.c, line 571 at r7](https://reviewable.io:443/reviews/pmem/pmemfile/291#-KtGgLpXNRL4QaOvgvtw:-KtGgLpY3nF0WDM7dIFz:b4h20z5) ([raw file](https://github.com/pmem/pmemfile/blob/14b4d9ac4ddd3d49e4a739599d67521d6ec6a2dc/src/libpmemfile/preload.c#L571)):*
> ```C
> 
> 	if (r < 0) {
> 		pmemfile_close(where->at_pool->pool, file);
> ```

+pool_acquire/release

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/291)*

GBuella:
Review status: 10 of 13 files reviewed at latest revision, 3 unresolved discussions.

---

*[src/libpmemfile/preload.c, line 571 at r7](https://reviewable.io:443/reviews/pmem/pmemfile/291#-KtGgLpXNRL4QaOvgvtw:-KtH-8B99TK3-ZuSEDEG:b-896fix) ([raw file](https://github.com/pmem/pmemfile/blob/14b4d9ac4ddd3d49e4a739599d67521d6ec6a2dc/src/libpmemfile/preload.c#L571)):*

+pool_acquire/release
</blockquote></details>

Done.

---

*[src/libpmemfile/vfd_table.c, line 405 at r2](https://reviewable.io:443/reviews/pmem/pmemfile/291#-KqbfDlP2bvWwh_vWWnk-r2-405:-KtH-A60itDr4QBJ1Aoj:b-896fix) ([raw file](https://github.com/pmem/pmemfile/blob/5b5bdd8a8b213029aadd681614d34fe108c0d0af/src/libpmemfile/vfd_table.c#L405)):*

assert(!entry->is_special_cwd_desc); ?
</blockquote></details>

Done.

---

*[src/libpmemfile/vfd_table.c, line 623 at r7](https://reviewable.io:443/reviews/pmem/pmemfile/291#-KtGfMNqyszfmylFyUbB:-KtH-BPD1FBjh1waDDRQ:b-896fix) ([raw file](https://github.com/pmem/pmemfile/blob/14b4d9ac4ddd3d49e4a739599d67521d6ec6a2dc/src/libpmemfile/vfd_table.c#L623)):*

"fd to keep"?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/291)*

marcinslusarz:
Reviewed 1 of 3 files at r7, 2 of 2 files at r8.
Review status: 13 of 14 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/291)*

marcinslusarz:
:lgtm:

---

Reviewed 1 of 1 files at r9.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/291#-:-KtHvTg2Bqgn0Ca5vWx6:bnfp4nl)*

PR #300, opened 2017-08-11
Use readlock in lseek, instead of writelock

GBuella:
This change is [Reviewable](https://reviewable.io/reviews/pmem/pmemfile/300)

### DIFF
--- a/src/libpmemfile-posix/inode.c
+++ b/src/libpmemfile-posix/inode.c
@@ -652,3 +652,43 @@ vinode_restore_on_abort(struct pmemfile_vinode *vinode)
 		vinode->blocks = NULL;
 	}
 }
+
+/*
+ * vinode_rdlock_with_block_tree - acquire read lock on a vinode instance,
+ * and rebuild the block_tree if needed.
+ *
+ * There are certain operations (e.g. read, lseek), which require the block
+ * tree to be valid, but do not modify the underlying file at all.
+ *
+ * If the block_tree doesn't exist we have to drop the lock we hold,
+ * take it in write mode (because other thread may want to do the same),
+ * check that it doesn't exist (another thread may already did that),
+ * drop the lock again, take it in read mode and check AGAIN (because
+ * another thread may have destroyed the block tree while we weren't
+ * holding the lock).
+ *
+ * On failure, returns a negative integer (the vinode->rwlock is not held).
+ * Returns zero on success (the caller can assume a read lock is held, and the
+ * block tree is valid).
+ */
+int
+vinode_rdlock_with_block_tree(PMEMfilepool *pfp, struct pmemfile_vinode *vinode)
+{
+	os_rwlock_rdlock(&vinode->rwlock);
+	while (!vinode->blocks) {
+		os_rwlock_unlock(&vinode->rwlock);
+		os_rwlock_wrlock(&vinode->rwlock);
+
+		int err = 0;
+		if (!vinode->blocks)
+			err = vinode_rebuild_block_tree(pfp, vinode);
+		os_rwlock_unlock(&vinode->rwlock);
+
+		if (err != 0)
+			return -err;
+
+		os_rwlock_rdlock(&vinode->rwlock);
+	}
+
+	return 0;
+}
--- a/src/libpmemfile-posix/inode.h
+++ b/src/libpmemfile-posix/inode.h
@@ -167,4 +167,6 @@ blockp_as_oid(struct pmemfile_block_desc *block)
 	return (TOID(struct pmemfile_block_desc))pmemobj_oid(block);
 }
 
+int vinode_rdlock_with_block_tree(PMEMfilepool *, struct pmemfile_vinode *);
+
 #endif
--- a/src/libpmemfile-posix/lseek.c
+++ b/src/libpmemfile-posix/lseek.c
@@ -49,11 +49,7 @@ static pmemfile_off_t
 lseek_seek_data(PMEMfilepool *pfp, struct pmemfile_vinode *vinode,
 		pmemfile_off_t offset, pmemfile_off_t fsize)
 {
-	if (vinode->blocks == NULL) {
-		int err = vinode_rebuild_block_tree(pfp, vinode);
-		if (err)
-			return err;
-	}
+	ASSERT(vinode->blocks != NULL);
 
 	struct pmemfile_block_desc *block =
 			find_closest_block(vinode, (uint64_t)offset);
@@ -84,11 +80,7 @@ static pmemfile_off_t
 lseek_seek_hole(PMEMfilepool *pfp, struct pmemfile_vinode *vinode,
 		pmemfile_off_t offset, pmemfile_off_t fsize)
 {
-	if (vinode->blocks == NULL) {
-		int err = vinode_rebuild_block_tree(pfp, vinode);
-		if (err)
-			return err;
-	}
+	ASSERT(vinode->blocks != NULL);
 
 	struct pmemfile_block_desc *block =
 			find_closest_block(vinode, (uint64_t)offset);
@@ -143,12 +135,19 @@ lseek_seek_data_or_hole(PMEMfilepool *pfp, struct pmemfile_vinode *vinode,
 		return -ENXIO;
 	}
 
+	/* Lock the vinode for read, rebuild block-tree if needed */
+	int r = vinode_rdlock_with_block_tree(pfp, vinode);
+	if (r != 0)
+		return r;
+
 	if (whence == PMEMFILE_SEEK_DATA) {
 		offset = lseek_seek_data(pfp, vinode, offset, fsize);
 	} else {
 		ASSERT(whence == PMEMFILE_SEEK_HOLE);
 		offset = lseek_seek_hole(pfp, vinode, offset, fsize);
 	}
+
+	os_rwlock_unlock(&vinode->rwlock);
 
 	if (offset > fsize)
 		offset = fsize;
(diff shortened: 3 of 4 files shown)


krzycz:
Reviewed 3 of 4 files at r1.
Review status: 3 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/libpmemfile-posix/inode.c, line 663 at r1](https://reviewable.io:443/reviews/pmem/pmemfile/300#-KrGHXeNjgQXlGzyydZh:-KrGHXeNjgQXlGzyydZi:bg5rx0m) ([raw file](https://github.com/pmem/pmemfile/blob/d344fed2e1d6f910b3b43cbafaceb3e4d9f0cd79/src/libpmemfile-posix/inode.c#L663)):*
> ```C
>  * tree to be valid, but do not modify the underlying file at all.
>  *
>  * If the block_tree doesn't exist we have to drop the lock we hold,
> ```

That puzzled me. Perhaps we need two locks per vinode?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/300)*

krzycz:
Reviewed 1 of 4 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/300)*

GBuella:

---

*[src/libpmemfile-posix/inode.c, line 663 at r1](https://reviewable.io:443/reviews/pmem/pmemfile/300#-KrGHXeNjgQXlGzyydZh:-KrGOuK4q47T3GSVYplf:bc8e8r2) ([raw file](https://github.com/pmem/pmemfile/blob/d344fed2e1d6f910b3b43cbafaceb3e4d9f0cd79/src/libpmemfile-posix/inode.c#L663)):*

That puzzled me. Perhaps we need two locks per vinode?
</blockquote></details>

If you mean a separate mutex inside each vinode called something like `block_tree_rebuild_mutex`, that actually sounds like a good idea.
Marcin, what do you think?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/300)*

marcinslusarz:
:lgtm:

---

Reviewed 4 of 4 files at r1.

---

*[src/libpmemfile-posix/inode.c, line 663 at r1](https://reviewable.io:443/reviews/pmem/pmemfile/300#-KrGHXeNjgQXlGzyydZh:-KrUwes6szwE_UrzICwq:b-okvo1g) ([raw file](https://github.com/pmem/pmemfile/blob/d344fed2e1d6f910b3b43cbafaceb3e4d9f0cd79/src/libpmemfile-posix/inode.c#L663)):*

If you mean a separate mutex inside each vinode called something like `block_tree_rebuild_mutex`, that actually sounds like a good idea.
Marcin, what do you think?
</blockquote></details>

It's not only about tree rebuild. Block tree needs to be consistent with blocks, so we need the same lock as is taken when writing (actually deallocating / allocating / initializing blocks). BTW, if we could fast & safely say whether metadata changed, very similar algo could be applied to overwriting.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/300#-:-KrUxd20iG8KKPdrqI8s:bnfp4nl)*

GBuella:

---

*[src/libpmemfile-posix/inode.c, line 663 at r1](https://reviewable.io:443/reviews/pmem/pmemfile/300#-KrGHXeNjgQXlGzyydZh:-KrV8FUKTWEGQXyslbgy:b-k7m8vd) ([raw file](https://github.com/pmem/pmemfile/blob/d344fed2e1d6f910b3b43cbafaceb3e4d9f0cd79/src/libpmemfile-posix/inode.c#L663)):*
> we need the same lock as is taken when writing
We would still hold the same lock as well -- the read-write-lock of the vinode, locked for reading. Noone should be able to do modifications while that.

Anyways, we expect it to be a very rare event to when the block tree is invalid, right?
Only on first access, and after aborted transactions I guess.
So this shouldn't matter much.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/300)*

marcinslusarz:
*[src/libpmemfile-posix/inode.c, line 663 at r1](https://reviewable.io:443/reviews/pmem/pmemfile/300#-KrGHXeNjgQXlGzyydZh:-KrVBUvIZXnoLKAYW1K6:bp40g2i) ([raw file](https://github.com/pmem/pmemfile/blob/d344fed2e1d6f910b3b43cbafaceb3e4d9f0cd79/src/libpmemfile-posix/inode.c#L663)):*

> we need the same lock as is taken when writing
We would still hold the same lock as well -- the read-write-lock of the vinode, locked for reading. Noone should be able to do modifications while that.

Anyways, we expect it to be a very rare event to when the block tree is invalid, right?
Only on first access, and after aborted transactions I guess.
So this shouldn't matter much.
</blockquote></details>

Yeah, that's a good idea (btw, you put first 2 sentences into the quote, which looks like I wrote that). The lock would only be needed for rebuilding, so we would have to be careful to update the pointer only when tree is fully populated.

 It's true it's rare event now, but it may become more frequent with process switching.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/300)*

sarahjelinek:
Reviewed 4 of 4 files at r1.
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

*[src/libpmemfile-posix/inode.c, line 678 at r1](https://reviewable.io:443/reviews/pmem/pmemfile/300#-KrXpfpCVaXq4ia9Yj33:-KrXpfpCVaXq4ia9Yj34:b-vra0h6) ([raw file](https://github.com/pmem/pmemfile/blob/d344fed2e1d6f910b3b43cbafaceb3e4d9f0cd79/src/libpmemfile-posix/inode.c#L678)):*
> ```C
> {
> 	os_rwlock_rdlock(&vinode->rwlock);
> 	while (!vinode->blocks) {
> ```

Why do you need a 'while' loop and then the subsequent if (!vnode->blocks)? I assume in case another thread gets in between dropping the read lock and taking the write lock and rebuilds the tree? 

Also, if the vinode->blocks != NULL then are we guaranteed at this point that this block cache is up to date? I assume so since we don't check for any other flag.

---

*[src/libpmemfile-posix/lseek.c, line 137 at r1](https://reviewable.io:443/reviews/pmem/pmemfile/300#-KrXq5lIzkQSWQYNHUgU:-KrXq5lIzkQSWQYNHUgV:b-bpzj61) ([raw file](https://github.com/pmem/pmemfile/blob/d344fed2e1d6f910b3b43cbafaceb3e4d9f0cd79/src/libpmemfile-posix/lseek.c#L137)):*
> ```C
> 		return -ENXIO;
> 	}
> ```

Add a comment here saying that if the vinode_rdlock_xxx succeeds the read lock will be held.  

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/300)*

GBuella:
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

*[src/libpmemfile-posix/inode.c, line 678 at r1](https://reviewable.io:443/reviews/pmem/pmemfile/300#-KrXpfpCVaXq4ia9Yj33:-KrXzf0V5mnvq0DFSoAA:b-9xv3nq) ([raw file](https://github.com/pmem/pmemfile/blob/d344fed2e1d6f910b3b43cbafaceb3e4d9f0cd79/src/libpmemfile-posix/inode.c#L678)):*

Why do you need a 'while' loop and then the subsequent if (!vnode->blocks)? I assume in case another thread gets in between dropping the read lock and taking the write lock and rebuilds the tree? 

Also, if the vinode->blocks != NULL then are we guaranteed at this point that this block cache is up to date? I assume so since we don't check for any other flag.
</blockquote></details>

I relocated this code section from read.c, originally written by Marcin.

See original discussion in the initial PR:
https://github.com/pmem/pmemfile/pull/95#discussion_r114783010

Well, I realize reading that does help a lot.
Yes, the while loop is needed, since some other might sneak in between the unlock + lock here.

Also, anywhere in libpmemfile-posix, the vinode->blocks pointer (and whatever it points to ) can only be modified while holding the vinode lock . Anyone anywhere holding that lock must leave that vinode->blocks pointer either in a valid state (pointing to a complete tree), or leave it as NULL before releasing the lock. Therefore, when we acquire the lock here, and see that the vinode->blocks pointer is not NULL, we assume things are OK.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/300)*

GBuella:
Review status: 2 of 4 files reviewed at latest revision, 2 unresolved discussions.

---

*[src/libpmemfile-posix/lseek.c, line 137 at r1](https://reviewable.io:443/reviews/pmem/pmemfile/300#-KrXq5lIzkQSWQYNHUgU:-Krf99m6CYxAb4roLdJK:b-896fix) ([raw file](https://github.com/pmem/pmemfile/blob/d344fed2e1d6f910b3b43cbafaceb3e4d9f0cd79/src/libpmemfile-posix/lseek.c#L137)):*

Add a comment here saying that if the vinode_rdlock_xxx succeeds the read lock will be held.  
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/300)*

sarahjelinek:
:lgtm:

---

Reviewed 2 of 2 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/300#-:-Krf9cHXmgVPIYmf-xW8:bnfp4nl)*

marcinslusarz:
Reviewed 2 of 2 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/300)*

PR #319, opened 2017-08-31
posix: use allocation classes

igchor:
This change is [Reviewable](https://reviewable.io/reviews/pmem/pmemfile/319)


GBuella:
Reviewed 1 of 29 files at r1.
Review status: 1 of 29 files reviewed at latest revision, 1 unresolved discussion, some commit checks broke.

---

*[src/libpmemfile-posix/blocks.c, line 61 at r1](https://reviewable.io:443/reviews/pmem/pmemfile/319#-Ksrxmsh2ry-6BKW-BE1:-Ksrxmsh2ry-6BKW-BE2:b-53zl12) ([raw file](https://github.com/pmem/pmemfile/blob/abe15eca53092a0ecaf686cac07c4ba29fb7df18/src/libpmemfile-posix/blocks.c#L61)):*
> ```C
> 
> struct pobj_alloc_class_desc pmemfile_posix_block =
> 	{ 0, 1000, POBJ_HEADER_NONE, CONST_BLOCK_ID};
> ```

If I interpret it right, this is the new way to allow us to set a fixed block size, right?
Just an idea, this could be a bit simpler I think:

The `data_blocks` array can be a null terminated array (e.g. the last element has unit_size==0)
If a user sets fix block size using an environment variable, then we just need to update the first element of this array, and make the second element be the terminator.

So we would need to look at only the `data_blocks` array, instead of checking pmemfile_posix_block each time. Something like:
```
for (struct pobj_alloc_class_desc *desc = data_blocks; desc->unit_size !=0; ++desc) {
   ...
}
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/319)*

igchor:
Review status: 1 of 33 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/libpmemfile-posix/blocks.c, line 61 at r1](https://reviewable.io:443/reviews/pmem/pmemfile/319#-Ksrxmsh2ry-6BKW-BE1:-KssOu6nXb0UMpUtXfx4:b-896fix) ([raw file](https://github.com/pmem/pmemfile/blob/abe15eca53092a0ecaf686cac07c4ba29fb7df18/src/libpmemfile-posix/blocks.c#L61)):*

If I interpret it right, this is the new way to allow us to set a fixed block size, right?
Just an idea, this could be a bit simpler I think:

The `data_blocks` array can be a null terminated array (e.g. the last element has unit_size==0)
If a user sets fix block size using an environment variable, then we just need to update the first element of this array, and make the second element be the terminator.

So we would need to look at only the `data_blocks` array, instead of checking pmemfile_posix_block each time. Something like:
```
for (struct pobj_alloc_class_desc *desc = data_blocks; desc->unit_size !=0; ++desc) {
   ...
}
```
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/319)*

marcinslusarz:
First and second patch should be merged into one because at each commit it should be possible to build and run tests. (It's crucial for the ability to git bisect bugs)

Third and fourth should be merged into one or reordered - it's not possible to build pmemfile on Travis without the fourth patch.

marcinslusarz:
Reviewed 21 of 29 files at r1, 6 of 9 files at r2.
Review status: 27 of 34 files reviewed at latest revision, 12 unresolved discussions, some commit checks failed.

---

*[src/libpmemfile-posix/blocks.c, line 58 at r3](https://reviewable.io:443/reviews/pmem/pmemfile/319#-KtX90IOVn6RZ7tQ-ODk:-KtX90IPCq54Dr9zitOI:b5lgu0v) ([raw file](https://github.com/pmem/pmemfile/blob/993a7e137e399f558798a752051e21b60db8d72e/src/libpmemfile-posix/blocks.c#L58)):*
> ```C
> 
> void
> set_block_size_const(size_t size)
> ```

just set_block_size

---

*[src/libpmemfile-posix/blocks.h, line 48 at r3](https://reviewable.io:443/reviews/pmem/pmemfile/319#-KtWs96h8ZyzNJ_wy1Ty:-KtWs96h8ZyzNJ_wy1Tz:basawkz) ([raw file](https://github.com/pmem/pmemfile/blob/993a7e137e399f558798a752051e21b60db8d72e/src/libpmemfile-posix/blocks.h#L48)):*
> ```objc
> 
> /*
>  * block_alignment value is always equal to smallest
> ```

the smallest

---

*[src/libpmemfile-posix/blocks.h, line 50 at r3](https://reviewable.io:443/reviews/pmem/pmemfile/319#-KtWsAiiipDEHs5pmrvC:-KtWsAiiipDEHs5pmrvD:b-n003em) ([raw file](https://github.com/pmem/pmemfile/blob/993a7e137e399f558798a752051e21b60db8d72e/src/libpmemfile-posix/blocks.h#L50)):*
> ```objc
>  * block_alignment value is always equal to smallest
>  * block size, so it's either MIN_BLOCK_SIZE or
>  * pmemfile_posix_block size if PMEMFILE_BLOCK_SIZE
> ```

s/size//

---

*[src/libpmemfile-posix/blocks.h, line 79 at r3](https://reviewable.io:443/reviews/pmem/pmemfile/319#-KtWsCp_cjORv54UroPX:-KtWsCp_cjORv54UroPY:bzcbgc) ([raw file](https://github.com/pmem/pmemfile/blob/993a7e137e399f558798a752051e21b60db8d72e/src/libpmemfile-posix/blocks.h#L79)):*
> ```objc
> int initialize_alloc_classes(PMEMobjpool *pop);
> ```

too many empty lines

---

*[src/libpmemfile-posix/data.c, line 513 at r3](https://reviewable.io:443/reviews/pmem/pmemfile/319#-KtXPScAuPftbwVsC9vy:-KtXPScB42Jvrl6S032F:befqsy) ([raw file](https://github.com/pmem/pmemfile/blob/993a7e137e399f558798a752051e21b60db8d72e/src/libpmemfile-posix/data.c#L513)):*
> ```C
> 
> 			struct alloc_class_info info =
> 				data_block_info(size, MAX_BLOCK_SIZE);
> ```

Shouldn't second argument depend on "over"?

---

*[src/libpmemfile-posix/data.c, line 543 at r3](https://reviewable.io:443/reviews/pmem/pmemfile/319#-KtXPY-TIpBAf5vHFq2l:-KtXPY-TIpBAf5vHFq2m:b-7ropln) ([raw file](https://github.com/pmem/pmemfile/blob/993a7e137e399f558798a752051e21b60db8d72e/src/libpmemfile-posix/data.c#L543)):*
> ```C
> 
> 			struct alloc_class_info info =
> 				data_block_info(size, MAX_BLOCK_SIZE);
> ```

ditto

---

*[src/libpmemfile-posix/data.c, line 575 at r3](https://reviewable.io:443/reviews/pmem/pmemfile/319#-KtXPgqZA_kH8AwQSbCu:-KtXPgqZA_kH8AwQSbCv:bja6p0k) ([raw file](https://github.com/pmem/pmemfile/blob/3f1d2c48cf60f57f2c711a35dd5546437e91a70e/src/libpmemfile-posix/data.c#L575)):*
> ```C
> 			/* Are all those bytes needed? */
> 			if (hole_count > size)
> 				hole_count = size;
> ```

Why did you remove this?

---

*[src/libpmemfile-posix/inode.c, line 310 at r3](https://reviewable.io:443/reviews/pmem/pmemfile/319#-KtWsxnK9iJ-6Bx6DzKy:-KtWsxnK9iJ-6Bx6DzKz:b-b0mdih) ([raw file](https://github.com/pmem/pmemfile/blob/993a7e137e399f558798a752051e21b60db8d72e/src/libpmemfile-posix/inode.c#L310)):*
> ```C
> 
> 	if (inode_is_regular_file(inode))
> 		inode->file_data.blocks.length =
> ```

inode->file_data.blocks.version?

---

*[src/libpmemfile-posix/layout.h, line 218 at r3](https://reviewable.io:443/reviews/pmem/pmemfile/319#-KtWrKLqrGLVjBEl01vd:-KtWrKLqrGLVjBEl01ve:b-eiyvcw) ([raw file](https://github.com/pmem/pmemfile/blob/993a7e137e399f558798a752051e21b60db8d72e/src/libpmemfile-posix/layout.h#L218)):*
> ```objc
> 		+ 4  /* used */ \
> 		+ 8 /* padding */\
> 		+ sizeof(PMEMmutex) \
> ```

Please keep this list in the same order as fields in the structure.

---

*[src/libpmemfile-posix/pool.c, line 172 at r3](https://reviewable.io:443/reviews/pmem/pmemfile/319#-KtX9IUlCaqcKP5CF3m_:-KtX9IUlCaqcKP5CF3ma:b-ntvyk9) ([raw file](https://github.com/pmem/pmemfile/blob/993a7e137e399f558798a752051e21b60db8d72e/src/libpmemfile-posix/pool.c#L172)):*
> ```C
> 	}
> 
> 	if (initialize_alloc_classes(pfp->pop)) {
> ```

can we do it in initialize_super_block?

---

*[utils/docker/images/install-nvml.sh, line 37 at r3](https://reviewable.io:443/reviews/pmem/pmemfile/319#-KtXAX4lRwoS1mDr9d8i:-KtXAX4lRwoS1mDr9d8j:b-jc7pag) ([raw file](https://github.com/pmem/pmemfile/blob/993a7e137e399f558798a752051e21b60db8d72e/utils/docker/images/install-nvml.sh#L37)):*
> ```Shell
> #
> 
> git clone https://github.com/igchor/nvml.git
> ```

I think we should switch to pmemobj master, without this patch, but with smaller block size to not hit the memory corruption.
Also I think there's a benefit in keeping the compatibility with pmemobj 1.3 - missing allocation classes support should not result in failing to open the pool.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/319)*

igchor:
Review status: 27 of 34 files reviewed at latest revision, 11 unresolved discussions, some commit checks failed.

---

*[src/libpmemfile-posix/data.c, line 513 at r3](https://reviewable.io:443/reviews/pmem/pmemfile/319#-KtXPScAuPftbwVsC9vy:-Ktf_nw_M8WmVYuvrDCi:beobv33) ([raw file](https://github.com/pmem/pmemfile/blob/993a7e137e399f558798a752051e21b60db8d72e/src/libpmemfile-posix/data.c#L513)):*

Shouldn't second argument depend on "over"?
</blockquote></details>

Second argument is the biggest allowed block size, if it's MAX_BLOCK_SIZE, then there is no limit and data_block_info() will return the smallest block bigger or equal to "size". Do you want to be able to request bigger block size (depending on over)?

---

*[src/libpmemfile-posix/data.c, line 575 at r3](https://reviewable.io:443/reviews/pmem/pmemfile/319#-KtXPgqZA_kH8AwQSbCu:-KtffTQs6LZ3yj-w0Pol:b2sq9e6) ([raw file](https://github.com/pmem/pmemfile/blob/3f1d2c48cf60f57f2c711a35dd5546437e91a70e/src/libpmemfile-posix/data.c#L575)):*

Why did you remove this?
</blockquote></details>

Because in any case, we pass "size" as first argument to data_block_info() - we don't try to fill the entire hole.  E.g if size = 8KB and hole_count = 256KB: data_block_info(size, hole_count) will return block of size 16KB

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/319)*

krzycz:
General comment - reviewing the code that lacks comments/function briefs is not the most pleasant  things to do.

---

Reviewed 26 of 29 files at r1, 8 of 9 files at r2, 1 of 1 files at r3.
Review status: all files reviewed at latest revision, 17 unresolved discussions, some commit checks failed.

---

*[src/libpmemfile-posix/blocks.c, line 80 at r3](https://reviewable.io:443/reviews/pmem/pmemfile/319#-Ktk6GWlvdASqGsywKKS:-Ktk6GWlvdASqGsywKKT:b-dimrwb) ([raw file](https://github.com/pmem/pmemfile/blob/993a7e137e399f558798a752051e21b60db8d72e/src/libpmemfile-posix/blocks.c#L80)):*
> ```C
> 
> struct alloc_class_info
> metadata_block_info(void)
> ```

I don't get the idea of having a function that is a wrapper for another function that returns a **copy** of a static const structure, from which you only need a class_id (METADATA_ID) in most cases.
Why don't you define METADATA_ID in block.h and use it everywhere?

---

*[src/libpmemfile-posix/blocks.c, line 106 at r3](https://reviewable.io:443/reviews/pmem/pmemfile/319#-Ktk7eSZ3Tz8g_V0EZ7x:-Ktk7eSZ3Tz8g_V0EZ7y:b-jbojn7) ([raw file](https://github.com/pmem/pmemfile/blob/993a7e137e399f558798a752051e21b60db8d72e/src/libpmemfile-posix/blocks.c#L106)):*
> ```C
> {
> 	char query[30];
> 	struct pobj_alloc_class_desc desc = *ac;
> ```

Do you really need a local copy of a structure pointed by 'ac'?

---

*[src/libpmemfile-posix/dir.c, line 195 at r3](https://reviewable.io:443/reviews/pmem/pmemfile/319#-Ktk8D1qyhViK7OZo6tt:-Ktk8D1qyhViK7OZo6tu:bp0jvq9) ([raw file](https://github.com/pmem/pmemfile/blob/993a7e137e399f558798a752051e21b60db8d72e/src/libpmemfile-posix/dir.c#L195)):*
> ```C
> 			dir->next = TX_XALLOC(struct pmemfile_dir, info.size,
> 				POBJ_XALLOC_ZERO |
> 				POBJ_CLASS_ID(info.class_id));
> ```

Why not just `POBJ_CLASS_ID(METADATA_ID)`?

---

*[src/libpmemfile-posix/layout.h, line 98 at r3](https://reviewable.io:443/reviews/pmem/pmemfile/319#-Ktk3ugiJfLKK6TV3-DS:-Ktk3ugiJfLKK6TV3-DT:b-tm3tue) ([raw file](https://github.com/pmem/pmemfile/blob/993a7e137e399f558798a752051e21b60db8d72e/src/libpmemfile-posix/layout.h#L98)):*
> ```objc
> 
> 	/* padding / unused */
>     uint32_t padding_;
> ```

padding1/padding2 ?

---

*[src/libpmemfile-posix/pmemfile-posix.c, line 82 at r3](https://reviewable.io:443/reviews/pmem/pmemfile/319#-Ktk3RjhNO16HC5R9Ubn:-Ktk3RjhNO16HC5R9Ubo:b-1pbevd) ([raw file](https://github.com/pmem/pmemfile/blob/993a7e137e399f558798a752051e21b60db8d72e/src/libpmemfile-posix/pmemfile-posix.c#L82)):*
> ```C
> 	size_t pmemfile_posix_block_size = 0;
> 
> 	char *tmp = getenv("PMEMFILE_BLOCK_SIZE");
> ```

Why don't you use some meaningful names for these variables?
tmp => blk_size_env
tmpll => blk_size
...

---

*[utils/docker/images/install-nvml.sh, line 37 at r1](https://reviewable.io:443/reviews/pmem/pmemfile/319#-KssA5CyaoeFd8_GxuNj:-KssA5CzWrp4Bz9h2F-R:b-3udcye) ([raw file](https://github.com/pmem/pmemfile/blob/abe15eca53092a0ecaf686cac07c4ba29fb7df18/utils/docker/images/install-nvml.sh#L37)):*
> ```Shell
> #
> 
> git clone https://github.com/igchor/nvml.git
> ```

igchor ???

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/319#-:-Ktk8iA7Fvc0noUegmak:b-aclo4y)*

igchor:
Review status: 21 of 34 files reviewed at latest revision, 15 unresolved discussions.

---

*[src/libpmemfile-posix/blocks.c, line 58 at r3](https://reviewable.io:443/reviews/pmem/pmemfile/319#-KtX90IOVn6RZ7tQ-ODk:-KtjojT8Xo-kABx10keW:b-896fix) ([raw file](https://github.com/pmem/pmemfile/blob/993a7e137e399f558798a752051e21b60db8d72e/src/libpmemfile-posix/blocks.c#L58)):*

just set_block_size
</blockquote></details>

Done.

---

*[src/libpmemfile-posix/blocks.c, line 80 at r3](https://reviewable.io:443/reviews/pmem/pmemfile/319#-Ktk6GWlvdASqGsywKKS:-KtleWxDZihm0DIVto_K:b-q6ovt) ([raw file](https://github.com/pmem/pmemfile/blob/993a7e137e399f558798a752051e21b60db8d72e/src/libpmemfile-posix/blocks.c#L80)):*

I don't get the idea of having a function that is a wrapper for another function that returns a **copy** of a static const structure, from which you only need a class_id (METADATA_ID) in most cases.
Why don't you define METADATA_ID in block.h and use it everywhere?
</blockquote></details>

Right, I changed that and now I just return const pointer.  In new revision I don't use METADATA_ID directly to support nvml 1.3.

---

*[src/libpmemfile-posix/blocks.c, line 106 at r3](https://reviewable.io:443/reviews/pmem/pmemfile/319#-Ktk7eSZ3Tz8g_V0EZ7x:-KtlfRxgRp4ArJ0seQ8b:b-6pmpzj) ([raw file](https://github.com/pmem/pmemfile/blob/993a7e137e399f558798a752051e21b60db8d72e/src/libpmemfile-posix/blocks.c#L106)):*

Do you really need a local copy of a structure pointed by 'ac'?
</blockquote></details>

Right, that was unnecessary

---

*[src/libpmemfile-posix/blocks.h, line 48 at r3](https://reviewable.io:443/reviews/pmem/pmemfile/319#-KtWs96h8ZyzNJ_wy1Ty:-Ktjp3ZtRaODsC6A0V35:b-896fix) ([raw file](https://github.com/pmem/pmemfile/blob/993a7e137e399f558798a752051e21b60db8d72e/src/libpmemfile-posix/blocks.h#L48)):*

the smallest
</blockquote></details>

Done.

---

*[src/libpmemfile-posix/blocks.h, line 50 at r3](https://reviewable.io:443/reviews/pmem/pmemfile/319#-KtWsAiiipDEHs5pmrvC:-Ktk61clg-NWy9SJVqdD:b3kre) ([raw file](https://github.com/pmem/pmemfile/blob/993a7e137e399f558798a752051e21b60db8d72e/src/libpmemfile-posix/blocks.h#L50)):*

s/size//
</blockquote></details>

?

---

*[src/libpmemfile-posix/blocks.h, line 79 at r3](https://reviewable.io:443/reviews/pmem/pmemfile/319#-KtWsCp_cjORv54UroPX:-KtjpXrWPjhhcz3wzGob:b-896fix) ([raw file](https://github.com/pmem/pmemfile/blob/993a7e137e399f558798a752051e21b60db8d72e/src/libpmemfile-posix/blocks.h#L79)):*

too many empty lines
</blockquote></details>

Done.

---

*[src/libpmemfile-posix/data.c, line 543 at r3](https://reviewable.io:443/reviews/pmem/pmemfile/319#-KtXPY-TIpBAf5vHFq2l:-Ktk5qgtBPmDboZkMtFn:b32wyo) ([raw file](https://github.com/pmem/pmemfile/blob/993a7e137e399f558798a752051e21b60db8d72e/src/libpmemfile-posix/data.c#L543)):*

ditto
</blockquote></details>

up

---

*[src/libpmemfile-posix/dir.c, line 195 at r3](https://reviewable.io:443/reviews/pmem/pmemfile/319#-Ktk8D1qyhViK7OZo6tt:-Ktlfi9zy749Le7GY16J:b-x2o48n) ([raw file](https://github.com/pmem/pmemfile/blob/993a7e137e399f558798a752051e21b60db8d72e/src/libpmemfile-posix/dir.c#L195)):*

Why not just `POBJ_CLASS_ID(METADATA_ID)`?
</blockquote></details>

I still use metadata_block_info() to support nvml 1.3 (when allocation classes are not supported info->class_id will be 0)

---

*[src/libpmemfile-posix/inode.c, line 310 at r3](https://reviewable.io:443/reviews/pmem/pmemfile/319#-KtWsxnK9iJ-6Bx6DzKy:-KtjpyDrCaQ6jifPm_dF:b-896fix) ([raw file](https://github.com/pmem/pmemfile/blob/993a7e137e399f558798a752051e21b60db8d72e/src/libpmemfile-posix/inode.c#L310)):*

inode->file_data.blocks.version?
</blockquote></details>

Done.

---

*[src/libpmemfile-posix/layout.h, line 98 at r3](https://reviewable.io:443/reviews/pmem/pmemfile/319#-Ktk3ugiJfLKK6TV3-DS:-Ktlg0b3xm8kkH5YAGng:b-896fix) ([raw file](https://github.com/pmem/pmemfile/blob/993a7e137e399f558798a752051e21b60db8d72e/src/libpmemfile-posix/layout.h#L98)):*

padding1/padding2 ?
</blockquote></details>

Done.

---

*[src/libpmemfile-posix/layout.h, line 218 at r3](https://reviewable.io:443/reviews/pmem/pmemfile/319#-KtWrKLqrGLVjBEl01vd:-KtjqL54arpRpQWvKVmQ:b-896fix) ([raw file](https://github.com/pmem/pmemfile/blob/993a7e137e399f558798a752051e21b60db8d72e/src/libpmemfile-posix/layout.h#L218)):*

Please keep this list in the same order as fields in the structure.
</blockquote></details>

Done.

---

*[src/libpmemfile-posix/pmemfile-posix.c, line 82 at r3](https://reviewable.io:443/reviews/pmem/pmemfile/319#-Ktk3RjhNO16HC5R9Ubn:-Ktlg1PswkAMCJpQ3NHD:b-896fix) ([raw file](https://github.com/pmem/pmemfile/blob/993a7e137e399f558798a752051e21b60db8d72e/src/libpmemfile-posix/pmemfile-posix.c#L82)):*

Why don't you use some meaningful names for these variables?
tmp => blk_size_env
tmpll => blk_size
...
</blockquote></details>

Done.

---

*[src/libpmemfile-posix/pool.c, line 172 at r3](https://reviewable.io:443/reviews/pmem/pmemfile/319#-KtX9IUlCaqcKP5CF3m_:-Ktk5smBO9p5mNLQlpR-:b-896fix) ([raw file](https://github.com/pmem/pmemfile/blob/993a7e137e399f558798a752051e21b60db8d72e/src/libpmemfile-posix/pool.c#L172)):*

can we do it in initialize_super_block?
</blockquote></details>

Done.

---

*[utils/docker/images/install-nvml.sh, line 37 at r1](https://reviewable.io:443/reviews/pmem/pmemfile/319#-KssA5CyaoeFd8_GxuNj:-Ktlg2MuzeePLw0N8Tva:bdtd6by) ([raw file](https://github.com/pmem/pmemfile/blob/abe15eca53092a0ecaf686cac07c4ba29fb7df18/utils/docker/images/install-nvml.sh#L37)):*

igchor ???
</blockquote></details>

This commit was just to run this pull request, intended to remove this before merge

---

*[utils/docker/images/install-nvml.sh, line 37 at r3](https://reviewable.io:443/reviews/pmem/pmemfile/319#-KtXAX4lRwoS1mDr9d8i:-Ktk5uTGt8QQ__hG_2N0:b-896fix) ([raw file](https://github.com/pmem/pmemfile/blob/993a7e137e399f558798a752051e21b60db8d72e/utils/docker/images/install-nvml.sh#L37)):*

I think we should switch to pmemobj master, without this patch, but with smaller block size to not hit the memory corruption.
Also I think there's a benefit in keeping the compatibility with pmemobj 1.3 - missing allocation classes support should not result in failing to open the pool.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/319)*

igchor:
Review status: 21 of 34 files reviewed at latest revision, 15 unresolved discussions.

---

*[src/libpmemfile-posix/blocks.h, line 50 at r3](https://reviewable.io:443/reviews/pmem/pmemfile/319#-KtWsAiiipDEHs5pmrvC:-Ktlh1DNj7tk86YjjL7X:b-896fix) ([raw file](https://github.com/pmem/pmemfile/blob/993a7e137e399f558798a752051e21b60db8d72e/src/libpmemfile-posix/blocks.h#L50)):*
<details><summary><i>Previously, igchor wrote…</i></summary><blockquote>

?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/319)*

marcinslusarz:
Reviewed 9 of 13 files at r4.
Review status: 30 of 34 files reviewed at latest revision, 13 unresolved discussions.

---

*[src/libpmemfile-posix/blocks.c, line 53 at r5](https://reviewable.io:443/reviews/pmem/pmemfile/319#-KtlnCK7ZPNsXY9tJw7P:-KtlnCK7ZPNsXY9tJw7Q:b-djqozq) ([raw file](https://github.com/pmem/pmemfile/blob/b1d3668c6f2dbf17a61cebf87757b715b4513990/src/libpmemfile-posix/blocks.c#L53)):*
> ```C
> 	{ 256 * 1024,		16 },
> 	{ 2 * 1024 * 1024,	8 },
> 	{ 0 } // terminator
> ```

no C++ comments in C code please

---

*[src/libpmemfile-posix/blocks.c, line 100 at r5](https://reviewable.io:443/reviews/pmem/pmemfile/319#-KtljaLPPIS6I1Ra7LVA:-KtljaLPPIS6I1Ra7LVB:b-djqozq) ([raw file](https://github.com/pmem/pmemfile/blob/b1d3668c6f2dbf17a61cebf87757b715b4513990/src/libpmemfile-posix/blocks.c#L100)):*
> ```C
> }
> 
> #ifdef POBJ_CLASS_ID // if allocation classes are supported
> ```

no C++ comments in C code please

---

*[src/libpmemfile-posix/blocks.c, line 137 at r5](https://reviewable.io:443/reviews/pmem/pmemfile/319#-KtlltMXx9ZJy91K2NgK:-KtlltMXx9ZJy91K2NgL:b4i9d43) ([raw file](https://github.com/pmem/pmemfile/blob/b1d3668c6f2dbf17a61cebf87757b715b4513990/src/libpmemfile-posix/blocks.c#L137)):*
> ```C
> 	int ret = set_alloc_class(pop, &metadata_block, METADATA_ID);
> 	if (ret)
> 		return;
> ```

I think the decision about whether error should be ignored or not is on the caller side.

---

*[src/libpmemfile-posix/blocks.c, line 145 at r5](https://reviewable.io:443/reviews/pmem/pmemfile/319#-KtlnRCnObVE6yCGedXs:-KtlnRCnObVE6yCGedXt:b-7ropln) ([raw file](https://github.com/pmem/pmemfile/blob/b1d3668c6f2dbf17a61cebf87757b715b4513990/src/libpmemfile-posix/blocks.c#L145)):*
> ```C
> 		ret = set_alloc_class(pop, block, FIRST_BLOCK_ID + index);
> 		if (ret)
> 			return;
> ```

ditto

---

*[src/libpmemfile-posix/blocks.h, line 52 at r5](https://reviewable.io:443/reviews/pmem/pmemfile/319#-KtljWRK5O3_yMZPD7zp:-KtljWRK5O3_yMZPD7zq:b-djqozq) ([raw file](https://github.com/pmem/pmemfile/blob/b1d3668c6f2dbf17a61cebf87757b715b4513990/src/libpmemfile-posix/blocks.h#L52)):*
> ```objc
> 
> 
> // block_alignment value is always equal to the smallest block size
> ```

no C++ comments in C code please

---

*[src/libpmemfile-posix/data.c, line 530 at r5](https://reviewable.io:443/reviews/pmem/pmemfile/319#-KtljhhvrITnpuG_U_cp:-KtljhhvrITnpuG_U_cq:bdomcaj) ([raw file](https://github.com/pmem/pmemfile/blob/b1d3668c6f2dbf17a61cebf87757b715b4513990/src/libpmemfile-posix/data.c#L530)):*
> ```C
> 				count = (uint32_t)(first_offset - offset);
> 
> 				const struct pmem_block_info *info =
> ```

indentation is off here

---

*[utils/docker/images/install-nvml.sh, line 41 at r5](https://reviewable.io:443/reviews/pmem/pmemfile/319#-KtloGgCRCrZVTNGWZdQ:-KtloGgCRCrZVTNGWZdR:b1selj9) ([raw file](https://github.com/pmem/pmemfile/blob/b1d3668c6f2dbf17a61cebf87757b715b4513990/utils/docker/images/install-nvml.sh#L41)):*
> ```Shell
> 
> # checkout commit which introduces allocation classes
> git checkout 852860df2aed2e7ffb7f3aadc27be49d080ba81b
> ```

Please use recent commit id.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/319)*

GBuella:
Review status: 30 of 34 files reviewed at latest revision, 13 unresolved discussions.

---

*[src/libpmemfile-posix/blocks.c, line 53 at r5](https://reviewable.io:443/reviews/pmem/pmemfile/319#-KtlnCK7ZPNsXY9tJw7P:-Ktlpb66HT8ZI4xHfYRK:b5my89m) ([raw file](https://github.com/pmem/pmemfile/blob/b1d3668c6f2dbf17a61cebf87757b715b4513990/src/libpmemfile-posix/blocks.c#L53)):*

no C++ comments in C code please
</blockquote></details>

BTW, these are technically C comments, just sayin...

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/319)*

igchor:
Review status: 30 of 34 files reviewed at latest revision, 13 unresolved discussions.

---

*[src/libpmemfile-posix/blocks.c, line 53 at r5](https://reviewable.io:443/reviews/pmem/pmemfile/319#-KtlnCK7ZPNsXY9tJw7P:-KtoxdvQgPFKyQwTU8MU:b-896fix) ([raw file](https://github.com/pmem/pmemfile/blob/b1d3668c6f2dbf17a61cebf87757b715b4513990/src/libpmemfile-posix/blocks.c#L53)):*

BTW, these are technically C comments, just sayin...
</blockquote></details>

Done.

---

*[src/libpmemfile-posix/blocks.c, line 100 at r5](https://reviewable.io:443/reviews/pmem/pmemfile/319#-KtljaLPPIS6I1Ra7LVA:-KtoxPIoWbDbwKooEf2Y:b-896fix) ([raw file](https://github.com/pmem/pmemfile/blob/b1d3668c6f2dbf17a61cebf87757b715b4513990/src/libpmemfile-posix/blocks.c#L100)):*

no C++ comments in C code please
</blockquote></details>

Done.

---

*[src/libpmemfile-posix/blocks.c, line 137 at r5](https://reviewable.io:443/reviews/pmem/pmemfile/319#-KtlltMXx9ZJy91K2NgK:-KtovddE22F5rQTYvIxC:bjj52hz) ([raw file](https://github.com/pmem/pmemfile/blob/b1d3668c6f2dbf17a61cebf87757b715b4513990/src/libpmemfile-posix/blocks.c#L137)):*

I think the decision about whether error should be ignored or not is on the caller side.
</blockquote></details>

Ok, but set_alloc_class might return -1 when allocation classes are not supported and I think this should not result in error. we just don't use allocation classes then.

---

*[src/libpmemfile-posix/blocks.c, line 145 at r5](https://reviewable.io:443/reviews/pmem/pmemfile/319#-KtlnRCnObVE6yCGedXs:-Ktow5p67q00jR0DatfH:b32wyo) ([raw file](https://github.com/pmem/pmemfile/blob/b1d3668c6f2dbf17a61cebf87757b715b4513990/src/libpmemfile-posix/blocks.c#L145)):*

ditto
</blockquote></details>

up

---

*[src/libpmemfile-posix/blocks.h, line 52 at r5](https://reviewable.io:443/reviews/pmem/pmemfile/319#-KtljWRK5O3_yMZPD7zp:-KtoxQVqgv0fhvJYshFI:b-896fix) ([raw file](https://github.com/pmem/pmemfile/blob/b1d3668c6f2dbf17a61cebf87757b715b4513990/src/libpmemfile-posix/blocks.h#L52)):*

no C++ comments in C code please
</blockquote></details>

Done.

---

*[src/libpmemfile-posix/data.c, line 530 at r5](https://reviewable.io:443/reviews/pmem/pmemfile/319#-KtljhhvrITnpuG_U_cp:-KtoxbRjHapRYrKFFL_6:b-896fix) ([raw file](https://github.com/pmem/pmemfile/blob/b1d3668c6f2dbf17a61cebf87757b715b4513990/src/libpmemfile-posix/data.c#L530)):*

indentation is off here
</blockquote></details>

Done.

---

*[utils/docker/images/install-nvml.sh, line 41 at r5](https://reviewable.io:443/reviews/pmem/pmemfile/319#-KtloGgCRCrZVTNGWZdQ:-KtoxckQ20tiKZzbyDBm:b-896fix) ([raw file](https://github.com/pmem/pmemfile/blob/b1d3668c6f2dbf17a61cebf87757b715b4513990/utils/docker/images/install-nvml.sh#L41)):*

Please use recent commit id.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/319)*

marcinslusarz:
Review status: 30 of 34 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[src/libpmemfile-posix/blocks.c, line 137 at r5](https://reviewable.io:443/reviews/pmem/pmemfile/319#-KtlltMXx9ZJy91K2NgK:-KtpO5LFa3Qyi9Oy43D0:b-1vrctr) ([raw file](https://github.com/pmem/pmemfile/blob/b1d3668c6f2dbf17a61cebf87757b715b4513990/src/libpmemfile-posix/blocks.c#L137)):*
<details><summary><i>Previously, igchor wrote…</i></summary><blockquote>

Ok, but set_alloc_class might return -1 when allocation classes are not supported and I think this should not result in error. we just don't use allocation classes then.
</blockquote></details>

Just ignore it on the caller side. When we'll stop supporting 1.3 we will fail pool opening if allocation classes initialization failed.

BTW, please add error message when set_alloc_class fails.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/319)*

marcinslusarz:
Reviewed 2 of 4 files at r6, 1 of 1 files at r7.
Review status: 33 of 34 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/319)*

marcinslusarz:
Please rebase this PR. I'd like to see updated code coverage report using configuration from master.

krzycz:
Reviewed 9 of 13 files at r4, 3 of 4 files at r6, 1 of 1 files at r7.
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/319)*

igchor:
Done.

---

Review status: 22 of 34 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/libpmemfile-posix/blocks.c, line 137 at r5](https://reviewable.io:443/reviews/pmem/pmemfile/319#-KtlltMXx9ZJy91K2NgK:-Ktq37e6AI9k7Sz3UqYo:b-896fix) ([raw file](https://github.com/pmem/pmemfile/blob/b1d3668c6f2dbf17a61cebf87757b715b4513990/src/libpmemfile-posix/blocks.c#L137)):*

Just ignore it on the caller side. When we'll stop supporting 1.3 we will fail pool opening if allocation classes initialization failed.

BTW, please add error message when set_alloc_class fails.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/319#-:-Ktq356hB1o4gD5x6u2y:b-896fix)*

krzycz:
Reviewed 11 of 12 files at r8, 1 of 1 files at r9.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/319)*

marcinslusarz:
:lgtm:

---

Reviewed 11 of 12 files at r8, 1 of 1 files at r9.
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks pending.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/319#-:-Ktq7PolEcYu6_bi2Nq1:bnfp4nl)*

sarahjelinek:
Reviewed 12 of 29 files at r1, 6 of 13 files at r4, 1 of 4 files at r6, 10 of 12 files at r8, 1 of 1 files at r9.
Review status: all files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

*[src/libpmemfile-posix/dir.c, line 192 at r9](https://reviewable.io:443/reviews/pmem/pmemfile/319#-Ktwsv9ebXdKbSS3Hjd7:-Ktwsv9fOkq4t_zxHx4R:bfzail8) ([raw file](https://github.com/pmem/pmemfile/blob/f2be5bf646c81c7f75fbcbabf5573339f966632a/src/libpmemfile-posix/dir.c#L192)):*
> ```C
> 		if (!found && TOID_IS_NULL(dir->next)) {
> 			const struct pmem_block_info *info =
> 				metadata_block_info();
> ```

I don't think you need to split the line here. It won't be > 80 chars will it? IMO, when lines are split unnecessarily it is harder to read and it looks more like a conditional than an assignment. 

---

*[src/libpmemfile-posix/pmemfile-posix.c, line 87 at r9](https://reviewable.io:443/reviews/pmem/pmemfile/319#-KtwwjSWs9XNWO_5Sd5I:-KtwwjSWs9XNWO_5Sd5J:b-g1wm3o) ([raw file](https://github.com/pmem/pmemfile/blob/f2be5bf646c81c7f75fbcbabf5573339f966632a/src/libpmemfile-posix/pmemfile-posix.c#L87)):*
> ```C
> 		unsigned long long blk_size = strtoull(env, &end, 0);
> 		if (env[0] == '\0' || blk_size == ULLONG_MAX ||
> 				end[0] != '\0' || blk_size < MIN_BLOCK_SIZE) {
> ```

If there is no value associated with the PMEMFiLE_BLOCK_SIZE then env == NULL. I don't see a need to check for env[0] == '\0' here  again.  Also, end will be null if the call succeeds so I think checking for != NULL is fine, rather than checking the first character.

---

*[src/libpmemfile-posix/pool.c, line 87 at r9](https://reviewable.io:443/reviews/pmem/pmemfile/319#-KtwxhdWeV-_QmN2cq2V:-KtwxhdWeV-_QmN2cq2W:b-8nufc) ([raw file](https://github.com/pmem/pmemfile/blob/f2be5bf646c81c7f75fbcbabf5573339f966632a/src/libpmemfile-posix/pool.c#L87)):*
> ```C
> 	error = initialize_alloc_classes(pfp->pop);
> 	if (error) {
> 		/* allocation classes will not be used - ignore error */
> ```

Should you set error = 0 in this case if you are ignoring the error?  Because it will be set to non-zero and may result in a false failure below. 

---

*[src/libpmemfile-posix/pool.c, line 366 at r9](https://reviewable.io:443/reviews/pmem/pmemfile/319#-KtwyGMai2yKDlEx_L9Z:-KtwyGMai2yKDlEx_L9_:b-fjml9d) ([raw file](https://github.com/pmem/pmemfile/blob/f2be5bf646c81c7f75fbcbabf5573339f966632a/src/libpmemfile-posix/pool.c#L366)):*
> ```C
> 	error = initialize_alloc_classes(pfp->pop);
> 	if (error) {
> 		/* allocation classes will not be used - ignore error */
> ```

Set error = 0 here? 

---

*[src/libpmemfile-posix/stat.c, line 91 at r9](https://reviewable.io:443/reviews/pmem/pmemfile/319#-KtwyTL2miRs1NfPsZ1B:-KtwyTL2miRs1NfPsZ1C:b-8hos34) ([raw file](https://github.com/pmem/pmemfile/blob/f2be5bf646c81c7f75fbcbabf5573339f966632a/src/libpmemfile-posix/stat.c#L91)):*
> ```C
> 	} else if (inode_is_dir(inode)) {
> 		size_t sz = inode->size - sizeof(inode->file_data);
> 		blks = (pmemfile_blkcnt_t)(sz / 512);
> ```

If 512 is used in other places make it a #define. 

---

*[tests/posix/rw/rw.cpp, line 408 at r9](https://reviewable.io:443/reviews/pmem/pmemfile/319#-KtwzRz-LevQuP5k_wL_:-KtwzRz-LevQuP5k_wLa:b1lzhon) ([raw file](https://github.com/pmem/pmemfile/blob/f2be5bf646c81c7f75fbcbabf5573339f966632a/tests/posix/rw/rw.cpp#L408)):*
> ```cpp
> 
> 	if (env_block_size == 0x4000)
> 		EXPECT_TRUE(test_pmemfile_stats_match(pfp, 2, 0, 203, 12800));
> ```

Why did block size change to hex?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/319)*

igchor:
Review status: 33 of 34 files reviewed at latest revision, 6 unresolved discussions.

---

*[src/libpmemfile-posix/dir.c, line 192 at r9](https://reviewable.io:443/reviews/pmem/pmemfile/319#-Ktwsv9ebXdKbSS3Hjd7:-KtzHd5btJnayhV7Aom0:b-jg2fal) ([raw file](https://github.com/pmem/pmemfile/blob/f2be5bf646c81c7f75fbcbabf5573339f966632a/src/libpmemfile-posix/dir.c#L192)):*

I don't think you need to split the line here. It won't be > 80 chars will it? IMO, when lines are split unnecessarily it is harder to read and it looks more like a conditional than an assignment. 
</blockquote></details>

Actually it is > 80 chars, tabs are counted as 8 chars.

---

*[src/libpmemfile-posix/pmemfile-posix.c, line 87 at r9](https://reviewable.io:443/reviews/pmem/pmemfile/319#-KtwwjSWs9XNWO_5Sd5I:-KtzJ7LOr9Det4cTAPRc:bh7bq6c) ([raw file](https://github.com/pmem/pmemfile/blob/f2be5bf646c81c7f75fbcbabf5573339f966632a/src/libpmemfile-posix/pmemfile-posix.c#L87)):*

If there is no value associated with the PMEMFiLE_BLOCK_SIZE then env == NULL. I don't see a need to check for env[0] == '\0' here  again.  Also, end will be null if the call succeeds so I think checking for != NULL is fine, rather than checking the first character.
</blockquote></details>

This code existed before my patch so I may be wrong, but I think this check is to handle case when env is an empty string (it's not NULL in that case)

---

*[src/libpmemfile-posix/pool.c, line 87 at r9](https://reviewable.io:443/reviews/pmem/pmemfile/319#-KtwxhdWeV-_QmN2cq2V:-KtzF76uypyMJy-DrvCy:b-896fix) ([raw file](https://github.com/pmem/pmemfile/blob/f2be5bf646c81c7f75fbcbabf5573339f966632a/src/libpmemfile-posix/pool.c#L87)):*

Should you set error = 0 in this case if you are ignoring the error?  Because it will be set to non-zero and may result in a false failure below. 
</blockquote></details>

Done.

---

*[src/libpmemfile-posix/pool.c, line 366 at r9](https://reviewable.io:443/reviews/pmem/pmemfile/319#-KtwyGMai2yKDlEx_L9Z:-KtzF84ne_ZTtBkgQ0C5:b-896fix) ([raw file](https://github.com/pmem/pmemfile/blob/f2be5bf646c81c7f75fbcbabf5573339f966632a/src/libpmemfile-posix/pool.c#L366)):*

Set error = 0 here? 
</blockquote></details>

Done.

---

*[src/libpmemfile-posix/stat.c, line 91 at r9](https://reviewable.io:443/reviews/pmem/pmemfile/319#-KtwyTL2miRs1NfPsZ1B:-KtzFKiOFEcltJ2Tr9RC:bag9o63) ([raw file](https://github.com/pmem/pmemfile/blob/f2be5bf646c81c7f75fbcbabf5573339f966632a/src/libpmemfile-posix/stat.c#L91)):*

If 512 is used in other places make it a #define. 
</blockquote></details>

It is not used anywhere else, it's specific for that function.

---

*[tests/posix/rw/rw.cpp, line 408 at r9](https://reviewable.io:443/reviews/pmem/pmemfile/319#-KtwzRz-LevQuP5k_wL_:-KtzFRBhJ6u9FHunobyV:bq4j8ap) ([raw file](https://github.com/pmem/pmemfile/blob/f2be5bf646c81c7f75fbcbabf5573339f966632a/tests/posix/rw/rw.cpp#L408)):*

Why did block size change to hex?
</blockquote></details>

In some cases, block size was given in hex and in some other in decimal. I tried to change that so that all block sizes (in conditional statements at least) was using one convention (hex).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/319)*

krzycz:
Reviewed 1 of 1 files at r10.
Review status: all files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/319)*

GBuella:
Reviewed 1 of 13 files at r4, 1 of 12 files at r8.
Review status: all files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

*[src/libpmemfile-posix/dir.c, line 192 at r9](https://reviewable.io:443/reviews/pmem/pmemfile/319#-Ktwsv9ebXdKbSS3Hjd7:-KtzW02Sfh9Ku8aTpdX6:b-7vygd1) ([raw file](https://github.com/pmem/pmemfile/blob/f2be5bf646c81c7f75fbcbabf5573339f966632a/src/libpmemfile-posix/dir.c#L192)):*
<details><summary><i>Previously, igchor wrote…</i></summary><blockquote>

Actually it is > 80 chars, tabs are counted as 8 chars.
</blockquote></details>

Yes, reviewable show them as 4 chars, but cstyle counts them as 8. This would end at column 83.

---

*[src/libpmemfile-posix/pmemfile-posix.c, line 87 at r9](https://reviewable.io:443/reviews/pmem/pmemfile/319#-KtwwjSWs9XNWO_5Sd5I:-KtzXqTd97KA7CFK7B64:byw009a) ([raw file](https://github.com/pmem/pmemfile/blob/f2be5bf646c81c7f75fbcbabf5573339f966632a/src/libpmemfile-posix/pmemfile-posix.c#L87)):*
> If there is no value associated with the PMEMFiLE_BLOCK_SIZE then env == NULL.

Apparently getenv can return a pointer to an empty string, so yes, env[0] can be a null character. In that case strtoull can just return zero, without indicating any error.

Actually, this check could be moved to the `if` above, as in: `if(env && env[0] != '\0') {`

I just tried some scenarios with strtoull in glibc:

https://gist.github.com/GBuella/0ae002a871187a8421596f628ac8ab63

One should look at `blk_size == ULLONG_MAX`, as that indicates too large, or to small numbers.
One should look at *end, as that indicates wrong characters, e.g. if one sets `PMEMFILE_BLOCK_SIZE=16K` hoping for 16 kilobyte blocks.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/319)*

GBuella:
Review status: all files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

*[tests/posix/rw/rw.cpp, line 408 at r9](https://reviewable.io:443/reviews/pmem/pmemfile/319#-KtwzRz-LevQuP5k_wL_:-KtzdDq1jEV_CbFzTQUr:b-o2k6pl) ([raw file](https://github.com/pmem/pmemfile/blob/f2be5bf646c81c7f75fbcbabf5573339f966632a/tests/posix/rw/rw.cpp#L408)):*
<details><summary><i>Previously, igchor wrote…</i></summary><blockquote>

In some cases, block size was given in hex and in some other in decimal. I tried to change that so that all block sizes (in conditional statements at least) was using one convention (hex).
</blockquote></details>

And it is a bit more readable as well.
Especially if at some places we add a few special cases for other block sizes:
```
if (env_block_size == 0x1000) {
   ...
} else if (env_block_size == 0x6000) {
} else if (env_block_size == 0x11000) {
}
```

But since apparently we treat these values as related, we could add a name for this somewhere in the headers for tests, e.g.: `constexpr size_t test_blocksize = 0x4000; /* used for some tests with fix block size */`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/319)*

GBuella:
Review status: all files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[tests/posix/rw/rw.cpp, line 1210 at r10](https://reviewable.io:443/reviews/pmem/pmemfile/319#-KtzesMYojq-BpUHxLFi:-KtzesMYojq-BpUHxLFj:btnh43o) ([raw file](https://github.com/pmem/pmemfile/blob/4dbccabe4a2706915314a4b1e613251a66a2a94d/tests/posix/rw/rw.cpp#L1210)):*
> ```cpp
> 	ASSERT_EQ(pmemfile_lseek(pfp, f, 1, PMEMFILE_SEEK_DATA), 16384);
> 	ASSERT_EQ(pmemfile_lseek(pfp, f, 0, PMEMFILE_SEEK_DATA), 16384);
> 	ASSERT_EQ(pmemfile_lseek(pfp, f, 16382, PMEMFILE_SEEK_DATA), 16384);
> ```

And this could be `(4 * test_block_size) - 2`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/319)*

marcinslusarz:
Reviewed 1 of 1 files at r10.
Review status: all files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/319)*

marcinslusarz:
I think the issues Gabor pointed out are minor and can be fixed post-merge. Igor is on vacation this week, so he won't be able to fix it soon. This patch will conflict with what I'm working on.

marcinslusarz:
Minor fix for pmemobj 1.3 compatibility: https://github.com/marcinslusarz/pmemfile/commit/f6bb70a920d4af153980add0a56cda7dd12b759e

marcinslusarz:
And one more fix: https://github.com/marcinslusarz/pmemfile/commit/81a45589255061fc59f9ca0e6d13e166f0685769

sarahjelinek:
:lgtm:

---

Review status: all files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/319#-:-KuM33qAFyhd7HMQ_ulL:bnfp4nl)*

GBuella:
:lgtm:

---

Review status: all files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/319#-:-KuNXrOudqc8qB2Mz581:bnfp4nl)*

PR #322, opened 2017-09-01
posix: handle another case when we rely on undefined behavior in lseek

marcinslusarz:
SUMMARY: AddressSanitizer: undefined-behavior src/libpmemfile-posix/lseek.c:234:40 in
src/libpmemfile-posix/lseek.c:234:40: runtime error: negation of -9223372036854775808 cannot be represented in type 'pmemfile_off_t' (aka 'long'); cast to an unsigned type to negate this value to itself

---
This change is [Reviewable](https://reviewable.io/reviews/pmem/pmemfile/322)

### DIFF
--- a/src/libpmemfile-posix/lseek.c
+++ b/src/libpmemfile-posix/lseek.c
@@ -229,9 +229,15 @@ lseek_seek_data_or_hole(PMEMfilepool *pfp, struct pmemfile_vinode *vinode,
 static inline pmemfile_off_t
 add_off(size_t cur, pmemfile_off_t off)
 {
-	if (off >= 0)
-		return (pmemfile_off_t)(cur + (size_t)off);
-	return (pmemfile_off_t)(cur - (size_t)-off);
+	COMPILE_ERROR_ON(sizeof(cur) != sizeof(int64_t));
+	ASSERT(cur <= INT64_MAX);
+
+	cur += (size_t)off;
+
+	if (cur > INT64_MAX)
+		return -1;
+
+	return (pmemfile_off_t)cur;
 }
 
 /*


krzycz:
Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/322)*

src/libpmemfile-posix/lseek.c
```diff
@@ -231,6 +231,12 @@ add_off(size_t cur, pmemfile_off_t off)
 {
 	if (off >= 0)
 		return (pmemfile_off_t)(cur + (size_t)off);
+
+	/* -INT64_MIN can't be represented in pmemfile_off_t */
+	COMPILE_ERROR_ON(sizeof(pmemfile_off_t) != sizeof(int64_t));
+	if (off == INT64_MIN)
```
GBuella: I still say it would simpler/ cleaner:
```
size_t x = cur + (size_t)off;
if (x < cur || x > PMEMFILE_OFF_T_MAX)
   return -1;
else
   return (pmemfile_off_t)x;
```
No implementation defined, or undefined behaviour.

GBuella: Ah, right, except for the first cast.

src/libpmemfile-posix/lseek.c
```diff
@@ -231,6 +231,12 @@ add_off(size_t cur, pmemfile_off_t off)
 {
 	if (off >= 0)
 		return (pmemfile_off_t)(cur + (size_t)off);
```
GBuella: This original line might also try to cast INT64_MIN to size_t.

src/libpmemfile-posix/lseek.c
marcinslusarz: x can be smaller than cur when off is negative, so it's not that simple.

marcinslusarz: Ok, with just x < cur removed I think it's correct. I'll post new version in a minute.

GBuella:
:lgtm:

---

Reviewed 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/322#-:-Ksy7siRGNr_eDwn7ce_:bnfp4nl)*

PR #391, opened 2017-09-28
antool: fix match file of 'antool-analyze' test

ldorau:
Ten SyS_mmap calls come from ten pthread_create() calls in this file.
It happens that not all of them are detected, what causes sporadic
failures of this test. Make them optional, because detecting all of them
is not important.

---
This change is [Reviewable](https://reviewable.io/reviews/pmem/pmemfile/391)


marcinslusarz:
This looks suspiciously like a threading bug in vltrace...

ldorau:
@marcinslusarz
> This looks suspiciously like a threading bug in vltrace...

Yes ... The reason is: pthread_create() fails with the error "Resource temporarily unavailable".

ldorau:
pthread_create() may sometimes fail, so this test may too, so let's leave it as is.

PR #397, opened 2017-10-02
posix: replace ctree with direct offset mapping tree

igchor:
This change is [Reviewable](https://reviewable.io/reviews/pmem/pmemfile/397)


krzycz:
Reviewed 7 of 9 files at r1.
Review status: 7 of 9 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/libpmemfile-posix/offset_mapping.c, line 32 at r1](https://reviewable.io:443/reviews/pmem/pmemfile/397#-KvT_UYQXflQ7Fm_OZvh:-KvT_UYQXflQ7Fm_OZvi:bndr0xq) ([raw file](https://github.com/pmem/pmemfile/blob/4bfbf56e3f4607d6e76ba056cba5f8cc1a921447/src/libpmemfile-posix/offset_mapping.c#L32)):*
> ```C
>  * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
>  */
> #include <stdio.h>
> ```

any brief?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/397)*

GBuella:
Reviewed 8 of 9 files at r1.
Review status: 8 of 9 files reviewed at latest revision, 8 unresolved discussions.

---

*[src/libpmemfile-posix/CMakeLists.txt, line 94 at r1](https://reviewable.io:443/reviews/pmem/pmemfile/397#-KvTVlkya2T-S7CyOBTB:-KvTVlkya2T-S7CyOBTC:b-xipfn1) ([raw file](https://github.com/pmem/pmemfile/blob/4bfbf56e3f4607d6e76ba056cba5f8cc1a921447/src/libpmemfile-posix/CMakeLists.txt#L94)):*
> ```CMake
> 	out.h
> 	pool.h
> 	offset_mapping.h
> ```

Almost alphabetical order

---

*[src/libpmemfile-posix/data.c, line 67 at r1](https://reviewable.io:443/reviews/pmem/pmemfile/397#-KvTW0RjHwaCnF7R9ELu:-KvTW0RjHwaCnF7R9ELv:b-514sup) ([raw file](https://github.com/pmem/pmemfile/blob/4bfbf56e3f4607d6e76ba056cba5f8cc1a921447/src/libpmemfile-posix/data.c#L67)):*
> ```C
> find_last_block(const struct pmemfile_vinode *vinode)
> {
> 	return block_find_closest(vinode->blocks, UINT64_MAX);
> ```

Looks much better!

---

*[src/libpmemfile-posix/offset_mapping.c, line 41 at r1](https://reviewable.io:443/reviews/pmem/pmemfile/397#-KvTZ2hiC7r69Jy35jpT:-KvTZ2hj80MISK1SN4YY:b-6r10ob) ([raw file](https://github.com/pmem/pmemfile/blob/4bfbf56e3f4607d6e76ba056cba5f8cc1a921447/src/libpmemfile-posix/offset_mapping.c#L41)):*
> ```C
> 
> #define N_CHILDREN_POW 4
> #define N_CHILDREN (1 << N_CHILDREN_POW)
> ```

So, this will be a tree, with a branching factor of 16, right?
Each level has an array of 16 entries, each of which can internal, or leaf, right?

---

*[src/libpmemfile-posix/offset_mapping.c, line 68 at r1](https://reviewable.io:443/reviews/pmem/pmemfile/397#-KvTWslb6jKa4OZ45ho3:-KvTWslb6jKa4OZ45ho4:b-lyuop8) ([raw file](https://github.com/pmem/pmemfile/blob/4bfbf56e3f4607d6e76ba056cba5f8cc1a921447/src/libpmemfile-posix/offset_mapping.c#L68)):*
> ```C
>  */
> COMPILE_ERROR_ON(RANGE(ARRAY_SIZE(range) - 1) == 0);
> COMPILE_ERROR_ON(RANGE(ARRAY_SIZE(range)) != 0);
> ```

Doesn't some compiler complain about that? Perhaps that warning is not enabled?
I mean, we know that is what we really want here, but poor compiler doesn't understand that, I would guess. 

---

*[src/libpmemfile-posix/offset_mapping.c, line 76 at r1](https://reviewable.io:443/reviews/pmem/pmemfile/397#-KvTXNfjh-vTp7EivHW9:-KvTXNfjh-vTp7EivHWA:bibf0kb) ([raw file](https://github.com/pmem/pmemfile/blob/4bfbf56e3f4607d6e76ba056cba5f8cc1a921447/src/libpmemfile-posix/offset_mapping.c#L76)):*
> ```C
> offset_map_new(PMEMfilepool *pfp)
> {
> 	struct offset_map *m = pf_calloc(1, sizeof(struct offset_map));
> ```

pf_calloc(1, sizeof(*m));

---

*[src/libpmemfile-posix/offset_mapping.c, line 160 at r1](https://reviewable.io:443/reviews/pmem/pmemfile/397#-KvTaLxOx1RD1XCUiyTZ:-KvTaLxOx1RD1XCUiyT_:b-tli3wj) ([raw file](https://github.com/pmem/pmemfile/blob/4bfbf56e3f4607d6e76ba056cba5f8cc1a921447/src/libpmemfile-posix/offset_mapping.c#L160)):*
> ```C
> 		children = entry->child;
> 		entry = &children[offset / range[level]];
> 		offset %= range[level];
> ```

BTW, we could just do the shift in this loop, instead of adding a lookup table.
`range >>= N_CHILDREN_POW`

Generally, this could achieve the same thing as that lookup table:
```
static uint64_t range(int n)
{
     return UINT64_C(1) << (n * N_CHILDREN_POW);
}
```
Unless that lookup table is going to refer to non-constant branching factors.

---

*[src/libpmemfile-posix/offset_mapping.h, line 56 at r1](https://reviewable.io:443/reviews/pmem/pmemfile/397#-KvTcOExro3DGsD9377L:-KvTcOExro3DGsD9377M:b-yin4px) ([raw file](https://github.com/pmem/pmemfile/blob/4bfbf56e3f4607d6e76ba056cba5f8cc1a921447/src/libpmemfile-posix/offset_mapping.h#L56)):*
> ```objc
> 	 */
> 
> 	bool internal;
> ```

In a future version this one bit of information could be merged into that pointers least significant bit (using uintptr_t instead, or a two uintptr_t bitfields with sizes 63 and 1).
Normally I don't like doing such ugly things, but in this case it can mean doubling the branching factor using the amounts allocated, and potentially making the tree less tall.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/397)*

igchor:
Review status: 6 of 9 files reviewed at latest revision, 8 unresolved discussions.

---

*[src/libpmemfile-posix/CMakeLists.txt, line 94 at r1](https://reviewable.io:443/reviews/pmem/pmemfile/397#-KvTVlkya2T-S7CyOBTB:-KvWEYTzlBzthO6DqrUv:b-896fix) ([raw file](https://github.com/pmem/pmemfile/blob/4bfbf56e3f4607d6e76ba056cba5f8cc1a921447/src/libpmemfile-posix/CMakeLists.txt#L94)):*

Almost alphabetical order
</blockquote></details>

Done.

---

*[src/libpmemfile-posix/offset_mapping.c, line 32 at r1](https://reviewable.io:443/reviews/pmem/pmemfile/397#-KvT_UYQXflQ7Fm_OZvh:-KvWEHFcGubDSGT3mOLc:b-896fix) ([raw file](https://github.com/pmem/pmemfile/blob/4bfbf56e3f4607d6e76ba056cba5f8cc1a921447/src/libpmemfile-posix/offset_mapping.c#L32)):*

any brief?
</blockquote></details>

Done.

---

*[src/libpmemfile-posix/offset_mapping.c, line 41 at r1](https://reviewable.io:443/reviews/pmem/pmemfile/397#-KvTZ2hiC7r69Jy35jpT:-KvW27fkGjtj5FkZJWrR:b-ht0ukm) ([raw file](https://github.com/pmem/pmemfile/blob/4bfbf56e3f4607d6e76ba056cba5f8cc1a921447/src/libpmemfile-posix/offset_mapping.c#L41)):*

So, this will be a tree, with a branching factor of 16, right?
Each level has an array of 16 entries, each of which can internal, or leaf, right?
</blockquote></details>

That's right.

---

*[src/libpmemfile-posix/offset_mapping.c, line 68 at r1](https://reviewable.io:443/reviews/pmem/pmemfile/397#-KvTWslb6jKa4OZ45ho3:-KvW2LzytlXFHQ0B7Idx:b9acn4u) ([raw file](https://github.com/pmem/pmemfile/blob/4bfbf56e3f4607d6e76ba056cba5f8cc1a921447/src/libpmemfile-posix/offset_mapping.c#L68)):*

Doesn't some compiler complain about that? Perhaps that warning is not enabled?
I mean, we know that is what we really want here, but poor compiler doesn't understand that, I would guess. 
</blockquote></details>

What warning would that be? I don't really have any other idea how to check whether last entry is max possible value.

---

*[src/libpmemfile-posix/offset_mapping.c, line 76 at r1](https://reviewable.io:443/reviews/pmem/pmemfile/397#-KvTXNfjh-vTp7EivHW9:-KvWEVQjgk02aQhtxv9D:b-896fix) ([raw file](https://github.com/pmem/pmemfile/blob/4bfbf56e3f4607d6e76ba056cba5f8cc1a921447/src/libpmemfile-posix/offset_mapping.c#L76)):*

pf_calloc(1, sizeof(*m));
</blockquote></details>

Done.

---

*[src/libpmemfile-posix/offset_mapping.c, line 160 at r1](https://reviewable.io:443/reviews/pmem/pmemfile/397#-KvTaLxOx1RD1XCUiyTZ:-KvW2p8uvMs-1Nm1IILs:b-ab844u) ([raw file](https://github.com/pmem/pmemfile/blob/4bfbf56e3f4607d6e76ba056cba5f8cc1a921447/src/libpmemfile-posix/offset_mapping.c#L160)):*

BTW, we could just do the shift in this loop, instead of adding a lookup table.
`range >>= N_CHILDREN_POW`

Generally, this could achieve the same thing as that lookup table:
```
static uint64_t range(int n)
{
     return UINT64_C(1) << (n * N_CHILDREN_POW);
}
```
Unless that lookup table is going to refer to non-constant branching factors.
</blockquote></details>

Well, actually this lookup table is there for performance reasons, I tested it with similar function for calculating range and there was some difference. On this plot:
http://10.91.71.254:8080/view/Performance/job/pmemfile_fork_performance_sqlite/plot/

build 24 was with lookup table and 25 with calculating range as you suggested.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/397)*

GBuella:
Reviewed 2 of 3 files at r2.
Review status: 8 of 9 files reviewed at latest revision, 5 unresolved discussions.

---

*[src/libpmemfile-posix/offset_mapping.c, line 68 at r1](https://reviewable.io:443/reviews/pmem/pmemfile/397#-KvTWslb6jKa4OZ45ho3:-KvWXJKwodXxlm_KG3mC:b-vzfjov) ([raw file](https://github.com/pmem/pmemfile/blob/4bfbf56e3f4607d6e76ba056cba5f8cc1a921447/src/libpmemfile-posix/offset_mapping.c#L68)):*
<details><summary><i>Previously, igchor wrote…</i></summary><blockquote>

What warning would that be? I don't really have any other idea how to check whether last entry is max possible value.
</blockquote></details>

I was thinking about this one:
```
$ cat shifting.c
unsigned long long x(unsigned long long arg)
{
	return arg << 65;
}

$ cc shifting.c -c 
shifting.c: In function ‘x’:
shifting.c:4:13: warning: left shift count >= width of type [-Wshift-count-overflow]
  return arg << 65;
             ^~
```
And I didn't even explicitly enable any warnings...
Probably something silences it in the COMPILE_ERROR_ON macro, I'm not sure.

---

*[src/libpmemfile-posix/offset_mapping.c, line 160 at r1](https://reviewable.io:443/reviews/pmem/pmemfile/397#-KvTaLxOx1RD1XCUiyTZ:-KvWZY-t_0xebJtdx6Ou:bl1diq9) ([raw file](https://github.com/pmem/pmemfile/blob/4bfbf56e3f4607d6e76ba056cba5f8cc1a921447/src/libpmemfile-posix/offset_mapping.c#L160)):*
<details><summary><i>Previously, igchor wrote…</i></summary><blockquote>

Well, actually this lookup table is there for performance reasons, I tested it with similar function for calculating range and there was some difference. On this plot:
http://10.91.71.254:8080/view/Performance/job/pmemfile_fork_performance_sqlite/plot/

build 24 was with lookup table and 25 with calculating range as you suggested.
</blockquote></details>

wow, that is hard to understand why, but if you are sure

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/397)*

GBuella:
Review status: 8 of 9 files reviewed at latest revision, 5 unresolved discussions.

---

*[src/libpmemfile-posix/offset_mapping.c, line 160 at r1](https://reviewable.io:443/reviews/pmem/pmemfile/397#-KvTaLxOx1RD1XCUiyTZ:-KvW_-6edXQ23nYLok-C:behq7qo) ([raw file](https://github.com/pmem/pmemfile/blob/4bfbf56e3f4607d6e76ba056cba5f8cc1a921447/src/libpmemfile-posix/offset_mapping.c#L160)):*

wow, that is hard to understand why, but if you are sure
</blockquote></details>

I mean, compare:
 * calculate a memory address in a lookup table, and load from the lookup table
 * do a shift

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/397)*

GBuella:
Review status: 8 of 9 files reviewed at latest revision, 5 unresolved discussions.

---

*[src/libpmemfile-posix/offset_mapping.c, line 160 at r1](https://reviewable.io:443/reviews/pmem/pmemfile/397#-KvTaLxOx1RD1XCUiyTZ:-KvWbEMz9Wz5rcxy0ZvI:bf93pc4) ([raw file](https://github.com/pmem/pmemfile/blob/4bfbf56e3f4607d6e76ba056cba5f8cc1a921447/src/libpmemfile-posix/offset_mapping.c#L160)):*

I mean, compare:
 * calculate a memory address in a lookup table, and load from the lookup table
 * do a shift
</blockquote></details>

Well, I looked at it, both compile to two instructions in the general case, and both can compile to a single instruction in this loop

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/397)*

igchor:
Review status: 8 of 9 files reviewed at latest revision, 5 unresolved discussions.

---

*[src/libpmemfile-posix/offset_mapping.c, line 160 at r1](https://reviewable.io:443/reviews/pmem/pmemfile/397#-KvTaLxOx1RD1XCUiyTZ:-KvWfdAb-DdKCmtA9ON7:b-a8lq0h) ([raw file](https://github.com/pmem/pmemfile/blob/4bfbf56e3f4607d6e76ba056cba5f8cc1a921447/src/libpmemfile-posix/offset_mapping.c#L160)):*

Well, I looked at it, both compile to two instructions in the general case, and both can compile to a single instruction in this loop
</blockquote></details>

Well, actually I tested it not using (range >>= N_CHILDREN) but function similar to your's:
static uint64_t range(int n)
{
    return MIN_BLOCK_SIZE << (N_CHILDREN * (n - 1));
}

But probably you're right, I should just do: range >>= N_CHILDREN_POW. I dind't use that previously because I wasn't sure if branching factor should be const for each level.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/397)*

igchor:
Review status: 7 of 9 files reviewed at latest revision, 5 unresolved discussions.

---

*[src/libpmemfile-posix/offset_mapping.c, line 68 at r1](https://reviewable.io:443/reviews/pmem/pmemfile/397#-KvTWslb6jKa4OZ45ho3:-KvWpmtoxXEPecW2lYby:b3b4srz) ([raw file](https://github.com/pmem/pmemfile/blob/4bfbf56e3f4607d6e76ba056cba5f8cc1a921447/src/libpmemfile-posix/offset_mapping.c#L68)):*

I was thinking about this one:
```
$ cat shifting.c
unsigned long long x(unsigned long long arg)
{
	return arg << 65;
}

$ cc shifting.c -c 
shifting.c: In function ‘x’:
shifting.c:4:13: warning: left shift count >= width of type [-Wshift-count-overflow]
  return arg << 65;
             ^~
```
And I didn't even explicitly enable any warnings...
Probably something silences it in the COMPILE_ERROR_ON macro, I'm not sure.
</blockquote></details>

Okay, it's not needed anymore.

---

*[src/libpmemfile-posix/offset_mapping.c, line 160 at r1](https://reviewable.io:443/reviews/pmem/pmemfile/397#-KvTaLxOx1RD1XCUiyTZ:-KvWpvP8-0UNRkbc53xr:b-896fix) ([raw file](https://github.com/pmem/pmemfile/blob/4bfbf56e3f4607d6e76ba056cba5f8cc1a921447/src/libpmemfile-posix/offset_mapping.c#L160)):*
<details><summary><i>Previously, igchor wrote…</i></summary><blockquote>

Well, actually I tested it not using (range >>= N_CHILDREN) but function similar to your's:
static uint64_t range(int n)
{
    return MIN_BLOCK_SIZE << (N_CHILDREN * (n - 1));
}

But probably you're right, I should just do: range >>= N_CHILDREN_POW. I dind't use that previously because I wasn't sure if branching factor should be const for each level.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/397)*

GBuella:
Reviewed 1 of 2 files at r3.
Review status: 8 of 9 files reviewed at latest revision, 3 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/397)*

marcinslusarz:
Reviewed 5 of 9 files at r1, 1 of 3 files at r2.
Review status: 8 of 9 files reviewed at latest revision, 15 unresolved discussions.

---

*[src/libpmemfile-posix/block_array.c, line 477 at r3](https://reviewable.io:443/reviews/pmem/pmemfile/397#-KvXFpbT93LeL9GCoAGm:-KvXFpbU17hiXztnHEZP:be1zowk) ([raw file](https://github.com/pmem/pmemfile/blob/2818437a2196fdb255a448496856df528e521d6a/src/libpmemfile-posix/block_array.c#L477)):*
> ```C
> 		remove_block(vinode->blocks, moving_block);
> 		relocate_block(pfp, block, moving_block);
> 		insert_block(vinode->blocks, block);
> ```

insert_block can fail, so I think error handling should stay

---

*[src/libpmemfile-posix/data.c, line 37 at r3](https://reviewable.io:443/reviews/pmem/pmemfile/397#-KvXGNLq7xI6aXUENExD:-KvXGNLq7xI6aXUENExE:b-7rwxzi) ([raw file](https://github.com/pmem/pmemfile/blob/2818437a2196fdb255a448496856df528e521d6a/src/libpmemfile-posix/data.c#L37)):*
> ```C
> #include "data.h"
> #include "out.h"
> #include "offset_mapping.h"
> ```

f < u

---

*[src/libpmemfile-posix/inode.h, line 42 at r3](https://reviewable.io:443/reviews/pmem/pmemfile/397#-KvXGXjP0eOqCsmo_wQ_:-KvXGXjQkLy4q4eUc5xK:b-7rwxzk) ([raw file](https://github.com/pmem/pmemfile/blob/2818437a2196fdb255a448496856df528e521d6a/src/libpmemfile-posix/inode.h#L42)):*
> ```objc
> #include "layout.h"
> #include "os_thread.h"
> #include "offset_mapping.h"
> ```

f < s

---

*[src/libpmemfile-posix/offset_mapping.c, line 169 at r3](https://reviewable.io:443/reviews/pmem/pmemfile/397#-KvXPDo7o520G5j24k2t:-KvXPDo7o520G5j24k2u:bbmzcw8) ([raw file](https://github.com/pmem/pmemfile/blob/2818437a2196fdb255a448496856df528e521d6a/src/libpmemfile-posix/offset_mapping.c#L169)):*
> ```C
> 	if (entry->child != NULL) {
> 		return entry->child;
> 	} else {
> ```

s/} else {// 

---

*[src/libpmemfile-posix/offset_mapping.c, line 218 at r3](https://reviewable.io:443/reviews/pmem/pmemfile/397#-KvXPYncF2uz15QdAC3y:-KvXPYncF2uz15QdAC3z:b-uy2kpg) ([raw file](https://github.com/pmem/pmemfile/blob/2818437a2196fdb255a448496856df528e521d6a/src/libpmemfile-posix/offset_mapping.c#L218)):*
> ```C
>  * frees memory used by 'child' if  all child entries are NULL
>  */
> ```

remove empty line please

---

*[src/libpmemfile-posix/offset_mapping.c, line 307 at r3](https://reviewable.io:443/reviews/pmem/pmemfile/397#-KvXPVXpUNaCOcJUjwst:-KvXPVXqqv4GIDUvanYU:bzcbgc) ([raw file](https://github.com/pmem/pmemfile/blob/2818437a2196fdb255a448496856df528e521d6a/src/libpmemfile-posix/offset_mapping.c#L307)):*
> ```C
> 	}
> ```

too many empty lines

---

*[src/libpmemfile-posix/offset_mapping.c, line 327 at r3](https://reviewable.io:443/reviews/pmem/pmemfile/397#-KvXPjZvoPJbR5QWVpD2:-KvXPjZwfwthVHxrD6kS:b-lj33yp) ([raw file](https://github.com/pmem/pmemfile/blob/2818437a2196fdb255a448496856df528e521d6a/src/libpmemfile-posix/offset_mapping.c#L327)):*
> ```C
> 	 * cleans up offset_map tree
> 	 * if at the top level only first entry is internal and not null
> 	 * it's children can be transfered one level up and height of
> ```

its, transferred

---

*[src/libpmemfile-posix/offset_mapping.c, line 328 at r3](https://reviewable.io:443/reviews/pmem/pmemfile/397#-KvXPoN420ClbZuN7Es2:-KvXPoN420ClbZuN7Es3:b-4jwkpr) ([raw file](https://github.com/pmem/pmemfile/blob/2818437a2196fdb255a448496856df528e521d6a/src/libpmemfile-posix/offset_mapping.c#L328)):*
> ```C
> 	 * if at the top level only first entry is internal and not null
> 	 * it's children can be transfered one level up and height of
> 	 * the tree can be decresed
> ```

decreased

---

*[src/libpmemfile-posix/offset_mapping.c, line 340 at r3](https://reviewable.io:443/reviews/pmem/pmemfile/397#-KvXPtjo8ExHBzwLsLT8:-KvXPtjo8ExHBzwLsLT9:b-r1yi76) ([raw file](https://github.com/pmem/pmemfile/blob/2818437a2196fdb255a448496856df528e521d6a/src/libpmemfile-posix/offset_mapping.c#L340)):*
> ```C
> 				return 0;
> 
> 			/* check if all entries excpet first are NULL */
> ```

except

---

*[src/libpmemfile-posix/offset_mapping.h, line 47 at r3](https://reviewable.io:443/reviews/pmem/pmemfile/397#-KvXQXpnKvYAViqrUe8T:-KvXQXpnKvYAViqrUe8U:b5bddwg) ([raw file](https://github.com/pmem/pmemfile/blob/2818437a2196fdb255a448496856df528e521d6a/src/libpmemfile-posix/offset_mapping.h#L47)):*
> ```objc
> 	 * child points to pmemfile_block_desc when internal == false
> 	 * or to offset_map_entry array otherwise
> 	 */
> ```

Please move this comment before child field. I'm not sure it's worth it, but you could use union for the child field...

---

*[src/libpmemfile-posix/offset_mapping.h, line 58 at r3](https://reviewable.io:443/reviews/pmem/pmemfile/397#-KvXQLhFFiL6ETHcrUFs:-KvXQLhGFouwUOOY2zA8:b-fy4ofd) ([raw file](https://github.com/pmem/pmemfile/blob/2818437a2196fdb255a448496856df528e521d6a/src/libpmemfile-posix/offset_mapping.h#L58)):*
> ```objc
> 	PMEMfilepool *pfp;
> 
> 	uint64_t range;
> ```

comment before this field please

---

*[src/libpmemfile-posix/offset_mapping.h, line 63 at r3](https://reviewable.io:443/reviews/pmem/pmemfile/397#-KvXQQutZjEhX5A7M8Es:-KvXQQutZjEhX5A7M8Et:bzcbgc) ([raw file](https://github.com/pmem/pmemfile/blob/2818437a2196fdb255a448496856df528e521d6a/src/libpmemfile-posix/offset_mapping.h#L63)):*
> ```objc
> struct pmemfile_block_desc;
> ```

too many empty lines

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/397)*

krzycz:
Reviewed 1 of 3 files at r2, 2 of 2 files at r3.
Review status: all files reviewed at latest revision, 14 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/397)*

igchor:
Review status: 4 of 9 files reviewed at latest revision, 14 unresolved discussions.

---

*[src/libpmemfile-posix/block_array.c, line 477 at r3](https://reviewable.io:443/reviews/pmem/pmemfile/397#-KvXFpbT93LeL9GCoAGm:-KvYZVfMqsvcuthqyp6J:b-896fix) ([raw file](https://github.com/pmem/pmemfile/blob/2818437a2196fdb255a448496856df528e521d6a/src/libpmemfile-posix/block_array.c#L477)):*

insert_block can fail, so I think error handling should stay
</blockquote></details>

Done.

---

*[src/libpmemfile-posix/data.c, line 37 at r3](https://reviewable.io:443/reviews/pmem/pmemfile/397#-KvXGNLq7xI6aXUENExD:-KvXl79md_Ub4T4qiQGd:b-896fix) ([raw file](https://github.com/pmem/pmemfile/blob/2818437a2196fdb255a448496856df528e521d6a/src/libpmemfile-posix/data.c#L37)):*

f < u
</blockquote></details>

Done.

---

*[src/libpmemfile-posix/inode.h, line 42 at r3](https://reviewable.io:443/reviews/pmem/pmemfile/397#-KvXGXjP0eOqCsmo_wQ_:-KvXlBOEuRu9QXaR9vdn:b-896fix) ([raw file](https://github.com/pmem/pmemfile/blob/2818437a2196fdb255a448496856df528e521d6a/src/libpmemfile-posix/inode.h#L42)):*

f < s
</blockquote></details>

Done.

---

*[src/libpmemfile-posix/offset_mapping.c, line 169 at r3](https://reviewable.io:443/reviews/pmem/pmemfile/397#-KvXPDo7o520G5j24k2t:-KvYZWiLawT3ydauaNHa:b-kz0wp2) ([raw file](https://github.com/pmem/pmemfile/blob/2818437a2196fdb255a448496856df528e521d6a/src/libpmemfile-posix/offset_mapping.c#L169)):*

s/} else {// 
</blockquote></details>

I'm not sure what you mean

---

*[src/libpmemfile-posix/offset_mapping.c, line 218 at r3](https://reviewable.io:443/reviews/pmem/pmemfile/397#-KvXPYncF2uz15QdAC3y:-KvXlhvK1wMwWhQ65lZC:b-896fix) ([raw file](https://github.com/pmem/pmemfile/blob/2818437a2196fdb255a448496856df528e521d6a/src/libpmemfile-posix/offset_mapping.c#L218)):*

remove empty line please
</blockquote></details>

Done.

---

*[src/libpmemfile-posix/offset_mapping.c, line 307 at r3](https://reviewable.io:443/reviews/pmem/pmemfile/397#-KvXPVXpUNaCOcJUjwst:-KvXlmYutzsOBuwHnonO:b-896fix) ([raw file](https://github.com/pmem/pmemfile/blob/2818437a2196fdb255a448496856df528e521d6a/src/libpmemfile-posix/offset_mapping.c#L307)):*

too many empty lines
</blockquote></details>

Done.

---

*[src/libpmemfile-posix/offset_mapping.c, line 327 at r3](https://reviewable.io:443/reviews/pmem/pmemfile/397#-KvXPjZvoPJbR5QWVpD2:-KvXlypJnF4tofbRyBig:b-896fix) ([raw file](https://github.com/pmem/pmemfile/blob/2818437a2196fdb255a448496856df528e521d6a/src/libpmemfile-posix/offset_mapping.c#L327)):*

its, transferred
</blockquote></details>

Done.

---

*[src/libpmemfile-posix/offset_mapping.c, line 328 at r3](https://reviewable.io:443/reviews/pmem/pmemfile/397#-KvXPoN420ClbZuN7Es2:-KvXlzGHnoXKkBFBgr1q:b-896fix) ([raw file](https://github.com/pmem/pmemfile/blob/2818437a2196fdb255a448496856df528e521d6a/src/libpmemfile-posix/offset_mapping.c#L328)):*

decreased
</blockquote></details>

Done.

---

*[src/libpmemfile-posix/offset_mapping.c, line 340 at r3](https://reviewable.io:443/reviews/pmem/pmemfile/397#-KvXPtjo8ExHBzwLsLT8:-KvXly0XRw9S4KpCXC3z:b-896fix) ([raw file](https://github.com/pmem/pmemfile/blob/2818437a2196fdb255a448496856df528e521d6a/src/libpmemfile-posix/offset_mapping.c#L340)):*

except
</blockquote></details>

Done.

---

*[src/libpmemfile-posix/offset_mapping.h, line 47 at r3](https://reviewable.io:443/reviews/pmem/pmemfile/397#-KvXQXpnKvYAViqrUe8T:-KvYV7WpPJRY6NL53xZB:b-sj7z6r) ([raw file](https://github.com/pmem/pmemfile/blob/2818437a2196fdb255a448496856df528e521d6a/src/libpmemfile-posix/offset_mapping.h#L47)):*

Please move this comment before child field. I'm not sure it's worth it, but you could use union for the child field...
</blockquote></details>

Done. I was also thinking about union but whenever we access child we also check/set 'internal' value so it's always clear whether child points to block or next level entries. However, if you think that adding union would make code more readable, I can do that.

---

*[src/libpmemfile-posix/offset_mapping.h, line 58 at r3](https://reviewable.io:443/reviews/pmem/pmemfile/397#-KvXQLhFFiL6ETHcrUFs:-KvYWyUACRQyP0-EDJ7s:b-896fix) ([raw file](https://github.com/pmem/pmemfile/blob/2818437a2196fdb255a448496856df528e521d6a/src/libpmemfile-posix/offset_mapping.h#L58)):*

comment before this field please
</blockquote></details>

Done.

---

*[src/libpmemfile-posix/offset_mapping.h, line 63 at r3](https://reviewable.io:443/reviews/pmem/pmemfile/397#-KvXQQutZjEhX5A7M8Es:-KvYYhzT6tn7sRLQUtOP:b-896fix) ([raw file](https://github.com/pmem/pmemfile/blob/2818437a2196fdb255a448496856df528e521d6a/src/libpmemfile-posix/offset_mapping.h#L63)):*

too many empty lines
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/397)*

krzycz:
Reviewed 5 of 5 files at r4.
Review status: all files reviewed at latest revision, 14 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/397)*

marcinslusarz:
Review status: all files reviewed at latest revision, 5 unresolved discussions.

---

*[src/libpmemfile-posix/offset_mapping.c, line 169 at r3](https://reviewable.io:443/reviews/pmem/pmemfile/397#-KvXPDo7o520G5j24k2t:-KvapdPM2tUOp0mAMAl0:b-8uoo41) ([raw file](https://github.com/pmem/pmemfile/blob/2818437a2196fdb255a448496856df528e521d6a/src/libpmemfile-posix/offset_mapping.c#L169)):*
<details><summary><i>Previously, igchor wrote…</i></summary><blockquote>

I'm not sure what you mean
</blockquote></details>

It's a sed expression. When one code path of "if" ends with "return" there's no point in "else" and indentation in the other path.

---

*[src/libpmemfile-posix/offset_mapping.h, line 47 at r3](https://reviewable.io:443/reviews/pmem/pmemfile/397#-KvXQXpnKvYAViqrUe8T:-Kvaq8BkPuVGcUwxidNe:bf0fd10) ([raw file](https://github.com/pmem/pmemfile/blob/2818437a2196fdb255a448496856df528e521d6a/src/libpmemfile-posix/offset_mapping.h#L47)):*
<details><summary><i>Previously, igchor wrote…</i></summary><blockquote>

Done. I was also thinking about union but whenever we access child we also check/set 'internal' value so it's always clear whether child points to block or next level entries. However, if you think that adding union would make code more readable, I can do that.
</blockquote></details>

I was thinking about type safety and documenting of possible types. void* can be anything.

---

*[src/libpmemfile-posix/offset_mapping.h, line 58 at r3](https://reviewable.io:443/reviews/pmem/pmemfile/397#-KvXQLhFFiL6ETHcrUFs:-KvaqVPcYIOHY6Ll2hQP:b-ydafi9) ([raw file](https://github.com/pmem/pmemfile/blob/2818437a2196fdb255a448496856df528e521d6a/src/libpmemfile-posix/offset_mapping.h#L58)):*
<details><summary><i>Previously, igchor wrote…</i></summary><blockquote>

Done.
</blockquote></details>

"range" is usually 2 values (start and length, or start and end), so please document what is stored in "range" or rename it to something clearer

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/397)*

igchor:
Review status: 7 of 9 files reviewed at latest revision, 3 unresolved discussions.

---

*[src/libpmemfile-posix/offset_mapping.c, line 169 at r3](https://reviewable.io:443/reviews/pmem/pmemfile/397#-KvXPDo7o520G5j24k2t:-KvbCKKmOmyAeBqinIk_:b-896fix) ([raw file](https://github.com/pmem/pmemfile/blob/2818437a2196fdb255a448496856df528e521d6a/src/libpmemfile-posix/offset_mapping.c#L169)):*

It's a sed expression. When one code path of "if" ends with "return" there's no point in "else" and indentation in the other path.
</blockquote></details>

Done.

---

*[src/libpmemfile-posix/offset_mapping.h, line 47 at r3](https://reviewable.io:443/reviews/pmem/pmemfile/397#-KvXQXpnKvYAViqrUe8T:-KvbHHFMx5j42kRp1Hko:b-8wciht) ([raw file](https://github.com/pmem/pmemfile/blob/2818437a2196fdb255a448496856df528e521d6a/src/libpmemfile-posix/offset_mapping.h#L47)):*

I was thinking about type safety and documenting of possible types. void* can be anything.
</blockquote></details>

Done. However, I'm not sure if 'data' is a good name for an union.

---

*[src/libpmemfile-posix/offset_mapping.h, line 58 at r3](https://reviewable.io:443/reviews/pmem/pmemfile/397#-KvXQLhFFiL6ETHcrUFs:-KvbE4195w1040XVS0ZJ:b-qyt96s) ([raw file](https://github.com/pmem/pmemfile/blob/2818437a2196fdb255a448496856df528e521d6a/src/libpmemfile-posix/offset_mapping.h#L58)):*

"range" is usually 2 values (start and length, or start and end), so please document what is stored in "range" or rename it to something clearer
</blockquote></details>

Ok, you're right, probably range_length will be a better name.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/397)*

GBuella:
Review status: 7 of 9 files reviewed at latest revision, 4 unresolved discussions.

---

*[src/libpmemfile-posix/offset_mapping.c, line 162 at r5](https://reviewable.io:443/reviews/pmem/pmemfile/397#-KvbZMn0XK_8H8qwpM2E:-KvbZMn1fMFPyVRiVTur:b-3f07ow) ([raw file](https://github.com/pmem/pmemfile/blob/76a3dfe80bd54695d6c1e358c428398e495a2cf8/src/libpmemfile-posix/offset_mapping.c#L162)):*
> ```C
> 		range >>= N_CHILDREN_POW;
> 		children = entry->data.children;
> 		entry = &children[offset / range];
> ```

<OFFTOPIC>
For the curious:
I realized the compiler can't do strength reduction on the division here, since divisor comes form a variable -> doesn't know it is a power of two.
I compiled this, and there is a `div` instruction in the middle of this loop.

I managed to get rid of the division by using this uglier form:
```
int bit_index = __builtin_ctzll(range); /* cuz I know it has a single bit set, but the compiler doesn't know */

while (entry->internal) {
    children = entry->data.children;
    bit_index -= N_CHILDREN_POW;
    entry = children + ((offset >> bit_index) & ((1 << N_CHILDREN_POW) - 1));
}
```
</OFFTOPIC>

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/397)*

sarahjelinek:
Reviewed 4 of 9 files at r1, 1 of 3 files at r2, 3 of 5 files at r4.
Review status: 7 of 9 files reviewed at latest revision, 5 unresolved discussions.

---

*[src/libpmemfile-posix/offset_mapping.c, line 40 at r5](https://reviewable.io:443/reviews/pmem/pmemfile/397#-KvcDUR5nzYFQF4UXqLU:-KvcDUR5nzYFQF4UXqLV:b-66lmeo) ([raw file](https://github.com/pmem/pmemfile/blob/76a3dfe80bd54695d6c1e358c428398e495a2cf8/src/libpmemfile-posix/offset_mapping.c#L40)):*
> ```C
>  * one block at this range, offset_map_entry holds a pointer to an array of
>  * next level offset_map_entries (with smaller range) - size of this array
>  * is const and equal to N_CHILDREN defiend below
> ```

s/defiend/defined

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/397)*

GBuella:
Reviewed 1 of 5 files at r4, 2 of 2 files at r5.
Review status: all files reviewed at latest revision, 9 unresolved discussions.

---

*[src/libpmemfile-posix/block_array.c, line 478 at r5](https://reviewable.io:443/reviews/pmem/pmemfile/397#-KvcMOkKX8oMocUFjgUR:-KvcMOkKX8oMocUFjgUS:b-6z8hy0) ([raw file](https://github.com/pmem/pmemfile/blob/76a3dfe80bd54695d6c1e358c428398e495a2cf8/src/libpmemfile-posix/block_array.c#L478)):*
> ```C
> 		relocate_block(pfp, block, moving_block);
> 		if (insert_block(vinode->blocks, block))
> 			pmemfile_tx_abort(errno);
> ```

If inserting a block, then aborting the transaction?

---

*[src/libpmemfile-posix/offset_mapping.c, line 60 at r5](https://reviewable.io:443/reviews/pmem/pmemfile/397#-KvcO-E3vZLuSBoJIMVl:-KvcO-E3vZLuSBoJIMVm:b3hnh0t) ([raw file](https://github.com/pmem/pmemfile/blob/76a3dfe80bd54695d6c1e358c428398e495a2cf8/src/libpmemfile-posix/offset_mapping.c#L60)):*
> ```C
>  * ---------------------------------------------------------------------------
>  *                     ^(2)            ^(2)    ...       ^(2)
>  * 16 entries will be updated (blocks covering offsets 240k - 496k)
> ```

I see this business of representing ranges with ranges of different overlapping ranges (sorry if I don't make any sense)
 is pretty complicated to implement, are there separate unit tests just for the tree? It might be useful.
Is it correct with silly corner cases?

---

*[src/libpmemfile-posix/offset_mapping.c, line 180 at r5](https://reviewable.io:443/reviews/pmem/pmemfile/397#-KvcPR4p50GNtq6BOkTh:-KvcPR4p50GNtq6BOkTi:b-2dwqu8) ([raw file](https://github.com/pmem/pmemfile/blob/76a3dfe80bd54695d6c1e358c428398e495a2cf8/src/libpmemfile-posix/offset_mapping.c#L180)):*
> ```C
> 	struct offset_map_entry *e;
> 
> 	/* look for block in entries with lower offset */
> ```

Another little diagram here would be cool 

---

*[src/libpmemfile-posix/offset_mapping.c, line 195 at r5](https://reviewable.io:443/reviews/pmem/pmemfile/397#-KvcPX99czTWHSHHTl1z:-KvcPX99czTWHSHHTl2-:b-w1how6) ([raw file](https://github.com/pmem/pmemfile/blob/76a3dfe80bd54695d6c1e358c428398e495a2cf8/src/libpmemfile-posix/offset_mapping.c#L195)):*
> ```C
> 	/*
> 	 * look for block in entries with higher offset,
> 	 * if found, then return previous block
> ```

And a little diagram here

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/397)*

GBuella:
FYI, ctree tests:
https://github.com/pmem/nvml/blob/master/src/test/obj_ctree/obj_ctree.c

igchor:
Review status: all files reviewed at latest revision, 9 unresolved discussions.

---

*[src/libpmemfile-posix/block_array.c, line 478 at r5](https://reviewable.io:443/reviews/pmem/pmemfile/397#-KvcMOkKX8oMocUFjgUR:-Kvf_le4vkRz5jUYcj_D:b2w07na) ([raw file](https://github.com/pmem/pmemfile/blob/76a3dfe80bd54695d6c1e358c428398e495a2cf8/src/libpmemfile-posix/block_array.c#L478)):*

If inserting a block, then aborting the transaction?
</blockquote></details>

If inserting a block fails, then aborting the transaction. It looked the same for ctree_insert.

---

*[src/libpmemfile-posix/offset_mapping.c, line 40 at r5](https://reviewable.io:443/reviews/pmem/pmemfile/397#-KvcDUR5nzYFQF4UXqLU:-Kvf_irr47NXkKhU05Jr:b-896fix) ([raw file](https://github.com/pmem/pmemfile/blob/76a3dfe80bd54695d6c1e358c428398e495a2cf8/src/libpmemfile-posix/offset_mapping.c#L40)):*

s/defiend/defined
</blockquote></details>

Done.

---

*[src/libpmemfile-posix/offset_mapping.c, line 60 at r5](https://reviewable.io:443/reviews/pmem/pmemfile/397#-KvcO-E3vZLuSBoJIMVl:-Kvfa37ucJoSl-DzHsTM:buft3oe) ([raw file](https://github.com/pmem/pmemfile/blob/76a3dfe80bd54695d6c1e358c428398e495a2cf8/src/libpmemfile-posix/offset_mapping.c#L60)):*

I see this business of representing ranges with ranges of different overlapping ranges (sorry if I don't make any sense)
 is pretty complicated to implement, are there separate unit tests just for the tree? It might be useful.
Is it correct with silly corner cases?

</blockquote></details>

I didn't have enough time to write proper tests, which could be put in the repo yet. But I will try to create some tests today.

---

*[src/libpmemfile-posix/offset_mapping.c, line 180 at r5](https://reviewable.io:443/reviews/pmem/pmemfile/397#-KvcPR4p50GNtq6BOkTh:-Kvfe-0KZkLvFmQlIVFT:bs7wumb) ([raw file](https://github.com/pmem/pmemfile/blob/76a3dfe80bd54695d6c1e358c428398e495a2cf8/src/libpmemfile-posix/offset_mapping.c#L180)):*

Another little diagram here would be cool 
</blockquote></details>

Done.

---

*[src/libpmemfile-posix/offset_mapping.c, line 195 at r5](https://reviewable.io:443/reviews/pmem/pmemfile/397#-KvcPX99czTWHSHHTl1z:-Kvfe31bvZBDMF5B2t4d:b-896fix) ([raw file](https://github.com/pmem/pmemfile/blob/76a3dfe80bd54695d6c1e358c428398e495a2cf8/src/libpmemfile-posix/offset_mapping.c#L195)):*

And a little diagram here
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/397)*

GBuella:
Review status: 8 of 9 files reviewed at latest revision, 9 unresolved discussions.

---

*[src/libpmemfile-posix/offset_mapping.c, line 219 at r6](https://reviewable.io:443/reviews/pmem/pmemfile/397#-Kvfk8cgOi5iAF91DzH9:-Kvfk8chUEV_xh5twld5:b-g4qoy1) ([raw file](https://github.com/pmem/pmemfile/blob/eb8f9fb09c42b82c66be5889236f19e59dbafc8e/src/libpmemfile-posix/offset_mapping.c#L219)):*
> ```C
> 	 *
> 	 * entry mapping offset 48k to block is NULL, and there is
> 	 * no non-NULL entry to the left, so in this case BBB->prev
> ```

So, is this supposed to find the first block?
But that `prev` pointer must be null in this case anyways, or not?
We only store initialized blocks here?

Also, there is this field in struct pmemfile_vinode:
https://github.com/pmem/pmemfile/blob/master/src/libpmemfile-posix/inode.h#L97

---

*[src/libpmemfile-posix/offset_mapping.c, line 227 at r6](https://reviewable.io:443/reviews/pmem/pmemfile/397#-Kvfl0Bcm9MVsvFAr4Cd:-Kvfl0Bcm9MVsvFAr4Ce:bqbq5iz) ([raw file](https://github.com/pmem/pmemfile/blob/eb8f9fb09c42b82c66be5889236f19e59dbafc8e/src/libpmemfile-posix/offset_mapping.c#L227)):*
> ```C
> 			e++;
> 		else if (!e->internal) {
> 			return PF_RW(m->pfp, e->data.block->prev);
> ```

So can this return non-null?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/397)*

GBuella:
Review status: 8 of 9 files reviewed at latest revision, 8 unresolved discussions.

---

*[src/libpmemfile-posix/block_array.c, line 478 at r5](https://reviewable.io:443/reviews/pmem/pmemfile/397#-KvcMOkKX8oMocUFjgUR:-KvflWDdcL5OGRGNu58_:b-88fs13) ([raw file](https://github.com/pmem/pmemfile/blob/76a3dfe80bd54695d6c1e358c428398e495a2cf8/src/libpmemfile-posix/block_array.c#L478)):*
<details><summary><i>Previously, igchor wrote…</i></summary><blockquote>

If inserting a block fails, then aborting the transaction. It looked the same for ctree_insert.
</blockquote></details>

Ye, I see it was the same confusing thing in the original version as well.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/397)*

GBuella:
Review status: 8 of 9 files reviewed at latest revision, 5 unresolved discussions.

---

*[src/libpmemfile-posix/offset_mapping.c, line 219 at r6](https://reviewable.io:443/reviews/pmem/pmemfile/397#-Kvfk8cgOi5iAF91DzH9:-KvfoPh0Pu_pP8RyIQuj:b1q1pva) ([raw file](https://github.com/pmem/pmemfile/blob/eb8f9fb09c42b82c66be5889236f19e59dbafc8e/src/libpmemfile-posix/offset_mapping.c#L219)):*

So, is this supposed to find the first block?
But that `prev` pointer must be null in this case anyways, or not?
We only store initialized blocks here?

Also, there is this field in struct pmemfile_vinode:
https://github.com/pmem/pmemfile/blob/master/src/libpmemfile-posix/inode.h#L97
</blockquote></details>

OK, now I understand it finds the first block in a range, not the first block in the whole file.

---

*[src/libpmemfile-posix/offset_mapping.c, line 227 at r6](https://reviewable.io:443/reviews/pmem/pmemfile/397#-Kvfl0Bcm9MVsvFAr4Cd:-KvfoXa1nY9WvnXBl05y:b-ypbv2x) ([raw file](https://github.com/pmem/pmemfile/blob/eb8f9fb09c42b82c66be5889236f19e59dbafc8e/src/libpmemfile-posix/offset_mapping.c#L227)):*

So can this return non-null?
</blockquote></details>

I get it now.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/397)*

krzycz:
Reviewed 1 of 1 files at r6.
Review status: all files reviewed at latest revision, 5 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/397)*

igchor:
To make offset_mapping test compile I had to add:
`#ifdef __cplusplus`
`extern "C" {`
`#endif`

to offset_mapping.h and also define COMPILE_ERROR_ON to do nothing with c++.

If you have any ideas how to make this test compile without doing that, I will change that.

---

Review status: 3 of 13 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/397#-:-KvnM6jcFvuVluBeXB4h:b-srxynw)*

GBuella:
Review status: 3 of 13 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[src/compiler_utils.h, line 69 at r8](https://reviewable.io:443/reviews/pmem/pmemfile/397#-KvnunGjYr7n-dEG8nn-:-KvnunGjYr7n-dEG8nn0:b-kz9xij) ([raw file](https://github.com/pmem/pmemfile/blob/a184058d51ebcc4149c99dc9e5741da6ce84c825/src/compiler_utils.h#L69)):*
> ```objc
> 
> #ifdef __cplusplus
> #define COMPILE_ERROR_ON(cond)
> ```

We do use C++11, so this could be:
`#define COMPILE_ERROR_ON(cond) static_assert(!cond, "assertion failure")` , or something like that.
Some people have problem with using static_assert in C, even though it is supported pretty much everywhere -- and the NVML team instead invented their own wheel with opposite semantics...

But really noone should object to using it in C++, since it does not count as an extension since C++11.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/397)*

GBuella:
Reviewed 3 of 5 files at r7, 7 of 7 files at r8.
Review status: all files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/397)*

krzycz:
Reviewed 3 of 5 files at r7, 4 of 7 files at r8.
Review status: all files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/397)*

GBuella:
Review status: 7 of 36 files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

*[src/libpmemfile-posix/offset_mapping.h, line 45 at r9](https://reviewable.io:443/reviews/pmem/pmemfile/397#-KwAlyWqaf35jn7x5P21:-KwAlyWroJu_1pgOyJB7:b-7bvidk) ([raw file](https://github.com/pmem/pmemfile/blob/df1e048c01f0fcd89896ccb03f95d18f3a2f68b4/src/libpmemfile-posix/offset_mapping.h#L45)):*
> ```objc
> 
> /* branching factor is 2^N_CHILDREN_POW */
> #define N_CHILDREN_POW 4
> ```

I think these two constants should not be in the header file, and test should not know about these details.
Also, perhaps some thing would be easier if this offset_mapping would just be a general mapping from 64 bit integers to other 64 bit integers.
The tests include C headers, which were not meant to be compiled as C++, but testing a general map of integers should be possible with out using `pmemfile_block_desc`, `pmemobj_oid`, and such in tests.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/397)*

igchor:
Review status: 7 of 36 files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[src/libpmemfile-posix/offset_mapping.c, line 162 at r5](https://reviewable.io:443/reviews/pmem/pmemfile/397#-KvbZMn0XK_8H8qwpM2E:-KwB0n8aLFgAbSORiSSp:b-djjkti) ([raw file](https://github.com/pmem/pmemfile/blob/76a3dfe80bd54695d6c1e358c428398e495a2cf8/src/libpmemfile-posix/offset_mapping.c#L162)):*

<OFFTOPIC>
For the curious:
I realized the compiler can't do strength reduction on the division here, since divisor comes form a variable -> doesn't know it is a power of two.
I compiled this, and there is a `div` instruction in the middle of this loop.

I managed to get rid of the division by using this uglier form:
```
int bit_index = __builtin_ctzll(range); /* cuz I know it has a single bit set, but the compiler doesn't know */

while (entry->internal) {
    children = entry->data.children;
    bit_index -= N_CHILDREN_POW;
    entry = children + ((offset >> bit_index) & ((1 << N_CHILDREN_POW) - 1));
}
```
</OFFTOPIC>
</blockquote></details>

Ok, I've used this idea.

---

*[src/libpmemfile-posix/offset_mapping.h, line 45 at r9](https://reviewable.io:443/reviews/pmem/pmemfile/397#-KwAlyWqaf35jn7x5P21:-KwAzwWjht0ia4-_MB-V:bvn1vcv) ([raw file](https://github.com/pmem/pmemfile/blob/df1e048c01f0fcd89896ccb03f95d18f3a2f68b4/src/libpmemfile-posix/offset_mapping.h#L45)):*

I think these two constants should not be in the header file, and test should not know about these details.
Also, perhaps some thing would be easier if this offset_mapping would just be a general mapping from 64 bit integers to other 64 bit integers.
The tests include C headers, which were not meant to be compiled as C++, but testing a general map of integers should be possible with out using `pmemfile_block_desc`, `pmemobj_oid`, and such in tests.
</blockquote></details>

Okay, I could write test which wouldn't use this constants but in this case it would be harder to test some corner cases. Also if my algorithm would just map integers I wouldn't be able to return block->prev as in line 233 in offset_mapping.c It could be done by the caller like this:

`struct pmemfile_block_desc *block = block_find_closest( offset );`
` if (block->offset > offset) block = PF_RW(pfp, block->prev);`

But this additional check could probably decrease performance.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/397)*

krzycz:
Reviewed 29 of 29 files at r9.
Review status: all files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/397)*

marcinslusarz:
Reviewed 1 of 5 files at r4, 1 of 5 files at r7, 1 of 7 files at r8, 5 of 33 files at r10.
Review status: 9 of 14 files reviewed at latest revision, 7 unresolved discussions.

---

*[tests/posix/CMakeLists.txt, line 167 at r10](https://reviewable.io:443/reviews/pmem/pmemfile/397#-Kwdj5xqnXst1OVAolVV:-Kwdj5xrdBpqCNkCncgF:btwar88) ([raw file](https://github.com/pmem/pmemfile/blob/a2e178b32fa3bc7713be81b372a880456111cdfe/tests/posix/CMakeLists.txt#L167)):*
> ```CMake
> 
> if(FAULT_INJECTION)
>     target_sources(file_offset_mapping PRIVATE
> ```

Please always use tabs in cmake files.

---

*[tests/posix/offset_mapping/offset_mapping.cpp, line 60 at r10](https://reviewable.io:443/reviews/pmem/pmemfile/397#-Kweh7dVZWrehLOv6mS0:-Kweh7dVZWrehLOv6mS1:bo5n4y2) ([raw file](https://github.com/pmem/pmemfile/blob/a2e178b32fa3bc7713be81b372a880456111cdfe/tests/posix/offset_mapping/offset_mapping.cpp#L60)):*
> ```cpp
> 		return 1;
> 
> 	uint64_t result = x;
> ```

If you would start from 1 instead of x you would not need to have a special case of y==0.

---

*[tests/posix/offset_mapping/offset_mapping_wrapper.c, line 1 at r10](https://reviewable.io:443/reviews/pmem/pmemfile/397#-KwegX-pIb813O_H1Z6K:-KwegX-pIb813O_H1Z6L:b-4w3l4i) ([raw file](https://github.com/pmem/pmemfile/blob/a2e178b32fa3bc7713be81b372a880456111cdfe/tests/posix/offset_mapping/offset_mapping_wrapper.c#L1)):*
> ```C
> #include "layout.h"
> ```

Missing copyright and license.

---

*[tests/posix/offset_mapping/offset_mapping_wrapper.h, line 1 at r10](https://reviewable.io:443/reviews/pmem/pmemfile/397#-KwegTeedF89gSMpBJrC:-KwegTefG_BsYOKmTnjT:b-4w3l4i) ([raw file](https://github.com/pmem/pmemfile/blob/a2e178b32fa3bc7713be81b372a880456111cdfe/tests/posix/offset_mapping/offset_mapping_wrapper.h#L1)):*
> ```objc
> #ifndef OFFSET_MAPPING_WRAPPER_H
> ```

Missing copyright and license.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/397)*

igchor:
Review status: 9 of 14 files reviewed at latest revision, 7 unresolved discussions.

---

*[tests/posix/CMakeLists.txt, line 167 at r10](https://reviewable.io:443/reviews/pmem/pmemfile/397#-Kwdj5xqnXst1OVAolVV:-Kwj8YgLDk11Exf_JSuj:b-896fix) ([raw file](https://github.com/pmem/pmemfile/blob/a2e178b32fa3bc7713be81b372a880456111cdfe/tests/posix/CMakeLists.txt#L167)):*

Please always use tabs in cmake files.
</blockquote></details>

Done.

---

*[tests/posix/offset_mapping/offset_mapping.cpp, line 60 at r10](https://reviewable.io:443/reviews/pmem/pmemfile/397#-Kweh7dVZWrehLOv6mS0:-Kwj9ESEK4jkun1rUlhW:b-896fix) ([raw file](https://github.com/pmem/pmemfile/blob/a2e178b32fa3bc7713be81b372a880456111cdfe/tests/posix/offset_mapping/offset_mapping.cpp#L60)):*

If you would start from 1 instead of x you would not need to have a special case of y==0.
</blockquote></details>

Done.

---

*[tests/posix/offset_mapping/offset_mapping_wrapper.c, line 1 at r10](https://reviewable.io:443/reviews/pmem/pmemfile/397#-KwegX-pIb813O_H1Z6K:-KwjCdmY-2RQPd8nfxFQ:b-896fix) ([raw file](https://github.com/pmem/pmemfile/blob/a2e178b32fa3bc7713be81b372a880456111cdfe/tests/posix/offset_mapping/offset_mapping_wrapper.c#L1)):*

Missing copyright and license.
</blockquote></details>

Done.

---

*[tests/posix/offset_mapping/offset_mapping_wrapper.h, line 1 at r10](https://reviewable.io:443/reviews/pmem/pmemfile/397#-KwegTeedF89gSMpBJrC:-KwjCf1zYlDXSNT8SE6C:b-896fix) ([raw file](https://github.com/pmem/pmemfile/blob/a2e178b32fa3bc7713be81b372a880456111cdfe/tests/posix/offset_mapping/offset_mapping_wrapper.h#L1)):*

Missing copyright and license.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/397)*

marcinslusarz:
Reviewed 4 of 4 files at r11.
Review status: 13 of 14 files reviewed at latest revision, 3 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/397)*

marcinslusarz:
:lgtm:

---

Reviewed 1 of 33 files at r10.
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/397#-:-Kx86rPMWA_dyIY5_WsK:bnfp4nl)*

krzycz:
:lgtm:

---

Reviewed 1 of 7 files at r8, 5 of 33 files at r10.
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/397#-:-KxKFqhuTfEUZbOeJzTR:bnfp4nl)*

krzycz:
Reviewed 1 of 33 files at r10, 1 of 4 files at r11.
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/397)*

GBuella:
:lgtm:

---

Reviewed 2 of 29 files at r9, 6 of 33 files at r10, 4 of 4 files at r11.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/397#-:-KxN1zUukwAjW8LaXTZc:bnfp4nl)*

PR #406, opened 2017-10-04
posix: implement fcntl(..., F_SETFL, ...)

marcinslusarz:
... and add required locking in F_GETFL.

---
This change is [Reviewable](https://reviewable.io/reviews/pmem/pmemfile/406)

### DIFF
--- a/src/libpmemfile-posix/fcntl.c
+++ b/src/libpmemfile-posix/fcntl.c
@@ -84,10 +84,15 @@ pmemfile_fcntl(PMEMfilepool *pfp, PMEMfile *file, int cmd, ...)
 		}
 		case PMEMFILE_F_GETFL:
 		{
-			if (file->flags & PFILE_PATH)
-				return PMEMFILE_O_PATH;
-
 			int ret = 0;
+
+			os_mutex_lock(&file->mutex);
+
+			if (file->flags & PFILE_PATH) {
+				ret = PMEMFILE_O_PATH;
+				goto fl_end;
+			}
+
 			ret |= PMEMFILE_O_LARGEFILE;
 			if (file->flags & PFILE_APPEND)
 				ret |= PMEMFILE_O_APPEND;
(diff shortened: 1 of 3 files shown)


src/libpmemfile-posix/fcntl.c
```diff
@@ -102,8 +107,61 @@ pmemfile_fcntl(PMEMfilepool *pfp, PMEMfile *file, int cmd, ...)
...
+				errno = EINVAL;
+				return -1;
+
+			}
+#undef SUPPORTED_FLAGS
+
+			os_mutex_lock(&file->mutex);
+
+			if (fl_flags & PMEMFILE_O_APPEND)
+				file->flags |= PFILE_APPEND;
```
GBuella: This changes are shared among dup'ed fds.
So, to be clear, these are the *file status flags*, shared by dup'ed fd's, and the *file descriptor flags* are not shared, and those should be stored in the libpmemfile layer. Right?
And the only *file descriptor flag* we know about is the close-on-exec flag, if I interpret things right.

So, `F_SETFD` should be handled in libpmemfile, and not passed on to libpmemfile-posix, I guess.

src/libpmemfile-posix/fcntl.c
```diff
@@ -102,8 +107,61 @@ pmemfile_fcntl(PMEMfilepool *pfp, PMEMfile *file, int cmd, ...)
...
+					PMEMFILE_O_TRUNC | PMEMFILE_O_SYNC |
+					PMEMFILE_O_DSYNC);
+
+			if (fl_flags & PMEMFILE_O_ASYNC) {
+				ERR("setting O_ASYNC is not supported");
+				errno = EINVAL;
+				return -1;
+			}
+
+#define SUPPORTED_FLAGS (PMEMFILE_O_APPEND | PMEMFILE_O_DIRECT | \
```
GBuella: How about `static const int supported_flags`?

src/libpmemfile-posix/fcntl.c
marcinslusarz: You are right, but it's not relevant to this patch :). This patch is about F_SETFL, not F_SETFD.

GBuella: When I looked at it, I just wasn't sure if these changes should be OK with dup, or not, but if you agree too, then OK.

src/libpmemfile-posix/fcntl.c
marcinslusarz: Sure. Done.

sarahjelinek:
:lgtm:

---

Reviewed 2 of 3 files at r1, 1 of 1 files at r2.
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/406#-:-KvcAx4DA31ajxpw-rae:bnfp4nl)*

PR #407, opened 2017-10-04
Fix pmemcheck integration!

marcinslusarz:
This change is [Reviewable](https://reviewable.io/reviews/pmem/pmemfile/407)

### DIFF
--- a/src/libpmemfile-posix/dir.c
+++ b/src/libpmemfile-posix/dir.c
@@ -191,6 +191,7 @@ inode_add_dirent(PMEMfilepool *pfp,
 			const struct pmem_block_info *info =
 				metadata_block_info();
 
+			TX_ADD_DIRECT(&dir->next);
 			dir->next = TX_XALLOC(struct pmemfile_dir, info->size,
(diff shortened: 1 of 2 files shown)


tests/posix/posix-helpers.cmake
```diff
@@ -95,6 +95,12 @@ function(execute name)
 		match(${BIN_DIR}/${TRACER}.err ${SRC_DIR}/${TRACER}.err.match)
 	endif()
 
+	# pmemcheck is a special snowflake and it doesn't set exit code when it
+	# detects an error, so we have to look at its output
+	if (${TRACER} STREQUAL pmemcheck AND NOT ERR MATCHES "ERROR SUMMARY: 0")
```
GBuella: BTW, there is a test property for that kind of thing:
https://cmake.org/cmake/help/v3.0/prop_test/FAIL_REGULAR_EXPRESSION.html
Can be used as in:
https://github.com/pmem/syscall_intercept/blob/master/test/CMakeLists.txt#L321

marcinslusarz: Indeed. I forgot about it.

PR #415, opened 2017-10-06
Very simple [f]statfs implementation and pointer validation mode.

marcinslusarz:
A lot of tests in pjdfstest verify file system limits by creating files with names of length around pathconf(_PC_NAME_MAX) and symlinks with target length around pathconf(_PC_PATH_MAX).
Pmemfile limits are the same as for ext4 (so those tests pass locally), but Travis uses other file system which has different limits and this makes those tests fail. To fix that the first patch implements [f]statfs, which just return pmemfile limits.

A lot of tests from LTP and pjdfstest verify file APIs by passing pointers to invalid memory. We can't handle this correctly and efficiently in userspace, because we don't have all the facilities kernel has.
We can't disable such tests, because they test multiple aspects of the API, not just how invalid pointers are handled.

This new pointer validation mode deals with invalid pointers by asking the kernel to dereference them. Current implementation is very slow and naive (write to a pipe), so it's not compiled-in by default and even when it's compiled in it's only enabled by environment variable. It can be improved later.
On Travis it's enabled only for LTP and pjdfstest jobs.

---
This change is [Reviewable](https://reviewable.io/reviews/pmem/pmemfile/415)


src/libpmemfile-posix/statfs.c
```diff
@@ -0,0 +1,63 @@
...
+
+	buf->f_type = PMEMFILE_SUPER_VERSION(0, 0);
+	buf->f_namelen = PMEMFILE_MAX_FILE_NAME;
+	buf->f_frsize = PMEMFILE_PATH_MAX;
+	buf->f_bsize = MIN_BLOCK_SIZE;
+	buf->f_flags = PMEMFILE_ST_NODEV | PMEMFILE_ST_NOEXEC |
+			PMEMFILE_ST_RELATIME | PMEMFILE_ST_SYNCHRONOUS;
+
+	/*
+	 * There's no way to get this info out of pmemobj. df command filters
```
GBuella: Is any github issue opened for pmemobj?
I'm sure it could provide a more useful guess than `1`, if the will is there.

src/libpmemfile/preload.c
```diff
@@ -1940,6 +1981,9 @@ dispatch_syscall_fd_first(long syscall_number,
 	case SYS_fstat:
 		return fd_first_pmemfile_fstat(arg0, arg1);
 
+	case SYS_fstatfs:
+		return hook_fstatfs(arg0, (struct statfs *)arg1);
```
GBuella: Or:
`return wrapper_pmemfile_statfs(arg->pool->pool, (struct pmemfile_statfs *)arg1);`

src/libpmemfile/preload.c
```diff
@@ -225,19 +226,182 @@ exit_with_msg(int ret, const char *msg)
...
+/*
+ * returns true when [buf, buf+len] is accessible
+ */
+static bool
+is_accessible(const void *ptr, size_t len)
+{
+	if (!validate_pointers)
+		return true;
+	if (!ptr || len == 0)
+		return true;
```
GBuella: So, if `ptr` is NULL, we say it is accessible????several question marks???

src/libpmemfile/preload.c
```diff
@@ -1919,46 +2197,89 @@ dispatch_syscall_fd_first(long syscall_number,
...
+		int ret;
+		if ((ret = verify_iovec(arg1, arg2)))
+			return ret;
 
-	case SYS_writev:
 		return fd_first_pmemfile_writev(arg0, arg1, arg2);
+	}
+
+	case SYS_read: {
+		if (!is_accessible((void *)arg1, (size_t)arg2))
```
GBuella: And this only checks for the memory region being readable, right?
While read's buffer should also be writable.
Not that I know of any similar easy-to-implement way of checking for that.

src/libpmemfile/preload.c
marcinslusarz: The code is correct, but naming is not. This function is supposed to say whether memory is safe to handle. NULL is not accessible, but all APIs should be prepared to handle this one specific value. Ideas for better name?

is_safe? is_valid? is_poisoned?

src/libpmemfile/preload.c
marcinslusarz: Yes, it only checks for memory to be readable. But it's enough to handle pjdfstest and LTP tests. It's not supposed to be anything more.

src/libpmemfile-posix/statfs.c
marcinslusarz: I filed pmem/issues#658

src/libpmemfile/preload.c
marcinslusarz: Done.

src/libpmemfile/preload.c
marcinslusarz: I moved this check out of is_accessible.

GBuella:
:lgtm:

---

Reviewed 14 of 17 files at r1, 4 of 4 files at r2.
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/415#-:-Kw4goKw4eUS7QK1Tncm:bnfp4nl)*

PR #431, opened 2017-10-14
posix: more tests & minor fixes

mslusarz:
This change is [Reviewable](https://reviewable.io/reviews/pmem/pmemfile/431)

### DIFF
--- a/src/libpmemfile-posix/CMakeLists.txt
+++ b/src/libpmemfile-posix/CMakeLists.txt
@@ -121,6 +121,11 @@ set(EXPORTED_SYMBOLS
 	pmemfile_mkdir
 	pmemfile_mkdirat
 	pmemfile_mknodat
+	pmemfile_mmap
+	pmemfile_mprotect
+	pmemfile_mremap
+	pmemfile_msync
+	pmemfile_munmap
 	pmemfile_open_parent
 	pmemfile_open
 	pmemfile_open_root
--- a/src/libpmemfile-posix/chmod.c
+++ b/src/libpmemfile-posix/chmod.c
@@ -65,7 +65,7 @@ vinode_chmod(PMEMfilepool *pfp, struct pmemfile_vinode *vinode,
 			pmemfile_tx_abort(EPERM);
 
 		struct pmemfile_time tm;
-		tx_get_current_time(&tm);
+		get_current_time(&tm);
 
 		inode_tx_set_ctime(inode, tm);
 
--- a/src/libpmemfile-posix/chmod.c
+++ b/src/libpmemfile-posix/chmod.c
@@ -50,22 +50,18 @@
  */
 static int
 vinode_chmod(PMEMfilepool *pfp, struct pmemfile_vinode *vinode,
-		pmemfile_mode_t mode)
+		pmemfile_mode_t mode, struct pmemfile_cred *cred)
 {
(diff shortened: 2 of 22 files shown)


krzycz:
Reviewed 16 of 16 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/431)*

GBuella:
Review status: all files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

*[src/libpmemfile-posix/utils.c, line 50 at r1](https://reviewable.io:443/reviews/pmem/pmemfile/431#-Kw_ffrtKD2gBzxUzv8Q:-Kw_ffrtKD2gBzxUzv8R:b-z0pfsp) ([raw file](https://github.com/pmem/pmemfile/blob/75aa0d6cb89e44963fd8585369c9bd9c1f1bec02/src/libpmemfile-posix/utils.c#L50)):*
> ```C
> 	pmemfile_timespec_t tm;
> 	if (clock_gettime(CLOCK_REALTIME, &tm)) {
> 		ERR("!clock_gettime");
> ```

Looks like overkill.
I'm thinking it is probably pointless to prepare for failures of clock_gettime. According to the docs, `CLOCK_REALTIME` should be supported everywhere, so the only possible error would be EFAULT.
The buffer is on the stack, declared in the line right before the clock_gettime call.
The fault injection code for this returns EINVAL, which would suggest here that CLOCK_REALTIME is supported sometimes, and not supported perhaps a few microseconds later -- seems like pointless to prepare for this scenario, I mean, if that happens, then maybe adding 2 + 2 will result in 5, and we don't check for that either.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/431)*

mslusarz:

---

*[src/libpmemfile-posix/utils.c, line 50 at r1](https://reviewable.io:443/reviews/pmem/pmemfile/431#-Kw_ffrtKD2gBzxUzv8Q:-Kx5DyflRPS6VGIRiufe:bv9j3t) ([raw file](https://github.com/pmem/pmemfile/blob/75aa0d6cb89e44963fd8585369c9bd9c1f1bec02/src/libpmemfile-posix/utils.c#L50)):*

Looks like overkill.
I'm thinking it is probably pointless to prepare for failures of clock_gettime. According to the docs, `CLOCK_REALTIME` should be supported everywhere, so the only possible error would be EFAULT.
The buffer is on the stack, declared in the line right before the clock_gettime call.
The fault injection code for this returns EINVAL, which would suggest here that CLOCK_REALTIME is supported sometimes, and not supported perhaps a few microseconds later -- seems like pointless to prepare for this scenario, I mean, if that happens, then maybe adding 2 + 2 will result in 5, and we don't check for that either.
</blockquote></details>

Ok. I handled this failure with an expectation that this function may need to be changed in the future. But I just realized none of those changes would mean that this function could fail...

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/431)*

GBuella:
:lgtm:

---

Reviewed 5 of 16 files at r1, 17 of 20 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pmem/pmemfile/431#-:-Kx7MEuDC_lqM_O7pqId:bnfp4nl)*

